PR #5, opened 2015-11-25
[#108440080] Add guides/GPG

dcarley:
Not everybody will have used GPG before, or recently. This guide is intended
to make it easier for people, including but not limited to new starters, to
setup their machines. We can link to it from other repos like
government-paas/credentials so that we don't have duplication.

It also gives us a bit of structure to the team manual. This doesn't
strictly fit under the "team" section. I expect that we'll have other short
guides like this in the future and maybe a top-level for Cloud Foundry
specific information.

I've mostly paraphrased these instructions from:

- https://github.com/alphagov/tsuru-ansible#setting-up-gpg-encrypted-vault-password-support

With the following changes:

- use long fingerprints instead of short fingerprints
- collapsed the installs into single commands
- omit the Linux `gpg-agent.conf` because it's not clear we *need* this and
  it can be added back again afterwards if we do
- omit `pwgen` which is specific to tsuru-ansible
- recommend 4096 instead of 2048 bit keys for increased safety
- the script is the one that I've been using rather than the one from
  ArchLinux[0] or somewhere else, which I think is slightly easier to
  understand

[0]: https://wiki.archlinux.org/index.php/GnuPG#gpg-agent

### DIFF
--- a/docs/guides/GPG.md
+++ b/docs/guides/GPG.md
@@ -1,0 +1,90 @@
+Some of our projects use [GPG][] to encrypt secrets so that they can only be
+decrypted by members of the team. These instructions may be useful if you
+haven't used GPG recently or ever at all.
+
+[GPG]: https://www.gnupg.org/
+
+# Install
+
+On Mac OS X:
+
+    brew install gpg gpg-agent
+
+On Ubuntu:
+
+    sudo apt-get update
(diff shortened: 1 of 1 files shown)


Jonty:
Seems reasonable.

PR #9, opened 2015-12-07
[#109029950] Add organisation to Docker Hub URL

dcarley:
To match the other URLs and help people locate the things that we own.


mtekel:
LG

PR #10, opened 2015-12-09
Start of useful documentation on debugging CF

Jonty:
This documents how to download a remote CF container that has
successfully staged but does not run correctly, allowing you to inspect
the contents and figure out why a user-uploaded app is not functioning.

### DIFF
--- a/docs/guides/CloudFoundry_debugging.md
+++ b/docs/guides/CloudFoundry_debugging.md
@@ -1,0 +1,32 @@
+This is a set of useful things that may help you debug CloudFoundry problems.
+
+## Downloading an app container
+If there are problems running an app then ````cf files```` will not function as the container must be running. An alternative is to grab the droplet image directly from CloudFoundry and unpack it locally. This can also be useful when you need to obtain a copy of a deployed app if you don't have the source.
+
+First, get the GUID of the app you want to investigate:
+````
+$ cf app APPNAME --guid
+5ffab66f-0be2-4a65-b9ff-40a902a05b50
+````
+
+Then use the ````cf curl```` wrapper to make an API call that retrieves the droplet and unpack it:
+````
+$ mkdir droplet && cd droplet
+$ cf curl /v2/apps/5ffab66f-0be2-4a65-b9ff-40a902a05b50/droplet/download > image.tar.gz
+$ tar -zxf image.tar.gz
+````
+
+**Note:** Tar will exit complaining about invalid gzip blocks - this is because the droplet has other data (such as a signature) appended to the end of the image. This isn't a problem, all the files for the image will be unpacked anyway.
+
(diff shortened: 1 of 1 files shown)


docs/guides/CloudFoundry_debugging.md
```diff
@@ -0,0 +1,32 @@
...
+First, get the GUID of the app you want to investigate:
+````
+$ cf app APPNAME --guid
+5ffab66f-0be2-4a65-b9ff-40a902a05b50
+````
+
+Then use the ````cf curl```` wrapper to make an API call that retrieves the droplet and unpack it:
+````
+$ mkdir droplet && cd droplet
+$ cf curl /v2/apps/5ffab66f-0be2-4a65-b9ff-40a902a05b50/droplet/download > image.tar.gz
```
keymon: What about using one line and avoid the temporary file.
```
f curl /v2/apps/5ffab66f-0be2-4a65-b9ff-40a902a05b50/droplet/download | tar -zvf -
```

Jonty:
@keymon I actually ended up re-extracting more than once after I'd changed a lot of the code and wanted a fresh copy, so it was useful to keep the droplet image around while working.

keymon:
I guess it could be downloaded again or change the command. 

But in that case I suggest that we avoid to keep the tgz file in the same path than the code. The reason is that the .tgz file can conflict with the buildpack scripts or with the app itself.

We can implemented it this way, for instance:

```
cf curl /v2/apps/5ffab66f-0be2-4a65-b9ff-40a902a05b50/droplet/download > image.tar.gz
mkdir droplet && cd droplet
tar -xvzf ../image.tar.gz
```

keymon:
Otherwise I suggest add this code to help copy and paste:

```
APP_NAME=eq-survey-runner
mkdir -p $APP_NAME
cf curl /v2/apps/$(cf app $APP_NAME --guid)/droplet/download | tar -xvzf -C $APP_NAME
```

docs/guides/CloudFoundry_debugging.md
```diff
@@ -0,0 +1,32 @@
...
+$ tar -zxf image.tar.gz
+````
+
+**Note:** Tar will exit complaining about invalid gzip blocks - this is because the droplet has other data (such as a signature) appended to the end of the image. This isn't a problem, all the files for the image will be unpacked anyway.
+
+You should now have everything used to run the app and all the build logs. ````app```` contains everything the user ````cf push````-ed:
```
keymon: I'm not sure, but I think that app contains everything that is created AFTER the application has been staged. It does not need to be exactly what the user pushed.

In my test I see the compiled .pyc and some .so files.

Shall we mention that?
I suggest we mention that.

keymon:
Apart of these two comments, LGTM.

I would add the comment saying that the `app/` dir contains the files after staging if @Jonty is fine with it and merge.

keymon:
LGTM, merging

PR #11, opened 2015-12-17
Update spruce guide now operand expressions PR is merged

bleach:
I was reading the spruce guide and noticed that the referenced PR was merged and released into spruce v0.13.0, so I updated the instructions to use go get to fetch the source instead of git.

I assumed we had no other reason to build from source.

I've tested that the OR operator example works after following the new installation steps.

### DIFF
--- a/docs/guides/spruce.md
+++ b/docs/guides/spruce.md
@@ -4,14 +4,10 @@ For more information, see the spruce readme and [examples](https://github.com/ge
 
 ## Installation
 
-For the time being we depend on the code in [this pull request](https://github.com/geofffranks/spruce/pull/69). This section will be updated once it is merged to master.
+You need to install spruce >= 0.13.0. This version added support for operand expressions.
 
 ```
-$ git clone https://github.com/geofffranks/spruce.git
-$ git checkout expr
-$ go get github.com/geofffranks/spruce # Fetches all the dependencies required for build step and installs them in $GOPATH
-$ go build
-$ mv spruce /usr/bin # Or any other directory in $PATH
+$ go get github.com/geofffranks/spruce # Builds spruce, including dependencies and puts it in $GOPATH/bin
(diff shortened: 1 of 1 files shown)


docs/guides/spruce.md
```diff
@@ -4,14 +4,10 @@ For more information, see the spruce readme and [examples](https://github.com/ge
...
-For the time being we depend on the code in [this pull request](https://github.com/geofffranks/spruce/pull/69). This section will be updated once it is merged to master.
-
 ```
-$ git clone https://github.com/geofffranks/spruce.git
-$ git checkout expr
 $ go get github.com/geofffranks/spruce # Fetches all the dependencies required for build step and installs them in $GOPATH
-$ go build
-$ mv spruce /usr/bin # Or any other directory in $PATH
+$ go build github.com/geofffranks/spruce
+$ go install github.com/geofffranks/spruce # installs spruce in $GOPATH/bin
```
alext: You can replace these 2 lines with a single `go get github.com/geofffranks/spruce`.

alext: Actually there's already a `go get` above, which should be all that's needed.  It might be worth a note that you can also use `go get -u` to update if you have an older version.

bleach: That's what I started out doing and I thought it hadn't worked, because I couldn't find the spruce binary. I was probably looking in the wrong place.

Thanks for that, I've removed the unnecessary steps.

bleach: On go get -u, I'm not sure it's going to work for people who previously installed spruce using the current master version of this guide, which asks them to copy the binary manually into their path.

I'll have a go at a note to cover that use case.

bleach:
I've updated based on your comments and rewritten commit history.

The upgrade section is possibly over the top now :)

PR #12, opened 2015-12-18
[#108528644] Architecture decision record - manifest management

timmow:
This is an idea that comes from [this blog
post](http://thinkrelevance.com/blog/2011/11/15/documenting-architecture-decisions)

It is useful to keep a record of decisions that have been made on the
project, and this seems like a reasonable format for it.

A [good example of this on another
project](https://github.com/openregister/presentation/blob/95ffd16deacd4ee2f4161f2d3af53805fa708fbe/doc/arch/adr-001-remove-kafka.md)

### DIFF
--- a/docs/arch/adr-001-manifest-management.md
+++ b/docs/arch/adr-001-manifest-management.md
@@ -1,0 +1,32 @@
+Context
+=======
+The alpha initially took the approach of starting with a vanilla set of cloud
+foundry manifests, and merging new values into it using spiff. This became
+difficult to reason about, and cf-release was forked because it was easier than
+overriding necessary values using spiff. However, the confusing spiff hierarchy
+remained.
+
+Decision
+========
+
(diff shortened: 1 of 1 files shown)


combor:
LGTM

combor:
Sorry, I realised after merging that I'm a little bit confused how it is rendered in mkdocs.  

For instance the section is called "Arch" and I would be good to rename it to "Architecture" and the subsection is called Adr 001 manifest management - not sure if that's intentional or it should be just called "Manifest management"

dcarley:
> Sorry, I realised after merging that I'm a little bit confused how it is rendered in mkdocs.
> 
> For instance the section is called "Arch" and I would be good to rename it to "Architecture" and the subsection is called Adr 001 manifest management - not sure if that's intentional or it should be just called "Manifest management"

np, I've tweaked it in #14

PR #13, opened 2015-12-22
[#109787160] Add link to process for handling AWS alerts

alext:


brettansley:
I've tested the link both logged in and logged out of Google and can confirm that I can access the document when logged in but not when logged out. I'm going to merge this pull request.

PR #15, opened 2016-01-04
[#109787160] Describe the types of AWS alerts

dcarley:
So that team members have some idea of when to look at the process document.
This description has been copied from the first paragraph of the doc.


bleach:
:+1: that makes it easier to see if you're in the right place.

PR #20, opened 2016-02-25
Add product section to orientation page

dcarley:
With links to blogs and videos that explain why we're here and what we've
been doing so far. We can add Rory's next blog post(s) to this as they are
published.


mtekel:
LG. builds, no broken formatting & all links work.

PR #21, opened 2016-03-07
Styleguides Straw man

alext:
## What

Add some team-specific styleguides.

This is very much a first pass to get something down. The items captured here came from alphagov/paas-cf#138. Some of them are still being discussed amongst the team.

## How to review

Discuss, debate etc.

## Who can review

Everyone on the team, although I propose that this should be left to @dcarley to merge as Tech Lead.

### DIFF
--- a/docs/styleguides/YAML.md
+++ b/docs/styleguides/YAML.md
@@ -1,0 +1,24 @@
+We produce lots of YAML. This sets out some guidelines for how this should be
+formatted.
+
+In general, follow the same formatting style as examples in the [YAML spec]().
+
+[YAML spec]: http://yaml.org/spec/1.2/spec.html
+
+# Indentation
+
+## Lists
+
+Lists should be indented 2 spaces beyond the parent key. eg:
+
+```
+foo:
+  - one
+  - two
+  - three
+```
+
+It's worth noting that the YAML spec differs from the style that concourse
+generated when returning pipelines. We've therefore had to pick one, and have
+chosen to stick with the YAML spec. This is also the indentation pattern that
+seems to be more commonly implemented by editors.
--- a/docs/styleguides/concourse_pipeline.md
+++ b/docs/styleguides/concourse_pipeline.md
@@ -5,6 +5,17 @@ applies.
 [Concourse]: https://concourse.ci
 [YAML styleguide]: YAML
 
+## Triggering patterns
+
+Most of our pipelines follow the following pattern for triggering jobs. This
+should be used unless there's a good reason not to.
+
+Pipelines start with an `init` job that bumps a `pipeline-trigger` semver
+resource. This resource is then used to trigger everything else througout the
+pipeline. This ensures that all of a pipeline runs, even if some of the
+jobs/tasks are a no-op. It also means there is a consistent way to initiate a
+pipeline run (where they are triggered manually).
+
 ## Tasks
 
 ### Naming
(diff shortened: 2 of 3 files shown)


docs/styleguides/YAML.md
```diff
@@ -0,0 +1,19 @@
...
+
+In general, follow the same formatting style as examples in the [YAML spec]().
+
+[YAML spec]: http://yaml.org/spec/1.2/spec.html
+
+# Indentation
+
+## Lists
+
+Lists should be indented 2 spaces beyond the parent key. eg:
```
saliceti: While I agree it looks nicer, I'm not in favour of it:
- [the yaml spec](http://www.yaml.org/spec/1.2/spec.html#id2799784) says “-” and indentation are redundant
- leaving indentation out is more practical for us as the lines end up smaller and we don't have to break them

alext: The spec says that it's permissible to not have the indentation, but all the examples in the spec use the extra indentation (eg http://yaml.org/spec/1.2/2009-07-21/spec.html#id2559116). This form of indentation is also what's assumed by editors (it's what vim's YAML plugin assumes at least), so if we do something different, we end up fighting the editor all the time.

saliceti: Which plugin are you using? Can you tweak its settings?

docs/styleguides/concourse_pipeline.md
```diff
@@ -0,0 +1,36 @@
...
+Pipelines start with an `init` job that bumps a `pipeline-trigger` semver
+resource. This resource is then used to trigger everything else througout the
+pipeline.
+
+## Tasks
+
+### Naming
+
+Tasks should be given names that are verbs. Inputs and outputs should be given
+names that are nouns. Inputs and outputs should not be given names that are the
```
dcarley: Could you include some examples of task and input/output names?

docs/styleguides/concourse_pipeline.md
```diff
@@ -0,0 +1,36 @@
...
+[YAML styleguide]: YAML
+
+## Triggering patterns
+
+Most of our pipelines follow the following pattern for triggering jobs. This
+should be used unless there's a good reason not to.
+
+Pipelines start with an `init` job that bumps a `pipeline-trigger` semver
+resource. This resource is then used to trigger everything else througout the
+pipeline.
```
dcarley: Can you expand on why we do this - to ensure that all of the pipeline runs even if a task/resource is a noop.

docs/styleguides/YAML.md
dcarley: I think we should describe the fact that the YAML spec and Concourse's implementation disagree and that we had to pick one.

I'm a bit wary of formatting guidelines that have to be enforced by human review, because we will occasionally miss things and people tend to take those kind of suggestions personally. Are we able to configure [yamllint](http://yamllint.readthedocs.org/en/latest/configuration.html) to do this for us instead?

docs/styleguides/concourse_pipeline.md
alext: Added an example.

docs/styleguides/concourse_pipeline.md
alext: Added more detail.

docs/styleguides/YAML.md
alext: I've added a bit more detail around this choice.

It would be possible to validate with with yamllint.  We'd need to add the following to the config file:
```yaml
rules:
  indentation:
    spaces: 2
    indent-sequences: yes
    check-multi-line-strings: no
```

At present this throws up lots of errors in paas-cf (mostly relating to the manifests as opposed to the pipelines). I'm not sure whether it's worth the noise to fix this...

PR #22, opened 2016-03-15
For comment only: Controlling production changes

bleach:
## What

This documents the approach we agreed to take to changing production while discussing a story to [create tags for production and staging](https://www.pivotaltracker.com/story/show/115318605)

It is a _proposal_, raised for visibility within the team and we won't decide anything until we have completed a spike that confirms that signing git tags is a good way of doing this.

## How to review

Check that it's an accurate and complete enough summary of the discussion we had to function as a useful decision record.

Do not merge (yet). We won't

## Who can review

@dcarley @Rory80Hz @mtekel @saliceti were present, so can confirm veracity.

Comment from others within the team is encouraged, either on the PR or in person.

### DIFF
--- a/docs/architecture_decision_records/ADR0XX-controlling-prod-releases.md
+++ b/docs/architecture_decision_records/ADR0XX-controlling-prod-releases.md
@@ -1,0 +1,85 @@
+
+Context
+=======
+
+At the time of writing we have two types of environment:
+
+- development environments: short lived environments belonging to inviduals and
+  pulling changes from branches
+- CI master environment: a persistent environment that applies changes as they
+  are pushed to the master branch in git
+
+We wish to add a production environment and only apply safe changes to it,
+meaning that:
+
+- the changeset has been [inspected by at least one other person before being
+  merged to master by pull
+request](https://gdstechnology.blog.gov.uk/2014/01/27/how-we-use-github/)
+- it has been through automated testing (currently Cloud Foundry smoke and
+  acceptance tests).
+
+We previously [spiked](https://www.pivotaltracker.com/story/show/109036472) how
+we might promote changes between concourse instances, decided that [git
+tags](https://git-scm.com/book/en/v2/Git-Basics-Tagging) were a reasonable
+solution and [amended](https://www.pivotaltracker.com/story/show/112497991) the
+git resource to support [creating annotated
+tags](https://github.com/concourse/git-resource/pull/32) and
+[reading](https://github.com/concourse/git-resource/pull/31) from git tags.
+
+There are advantages in having a low latency and automated path to production,
+which include:
+- the ability to apply urgent updates such as critical security updates quickly
+- making it practical to deploy smaller, more inspectable and reversible
+  changesets
+
+There were some residual things we were worried about:
+
+- We would like to ensure that the precise version transitions that production
+  will make are tested prior to production. 
+- By allowing concourse to push tags, we are also allowing its credentials to
+  be used to push actual commits, since github uses one privilege to allow
+  both.
+- If anyone gained access to one of our personal github accounts via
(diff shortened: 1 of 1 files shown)


docs/architecture_decision_records/ADR0XX-controlling-prod-releases.md
```diff
@@ -0,0 +1,85 @@
...
+  are pushed to the master branch in git
+
+We wish to add a production environment and only apply safe changes to it,
+meaning that:
+
+- the changeset has been [inspected by at least one other person before being
+  merged to master by pull
+request](https://gdstechnology.blog.gov.uk/2014/01/27/how-we-use-github/)
+- it has been through automated testing (currently Cloud Foundry smoke and
+  acceptance tests).
```
combor: We've got failure testing pipeline also, but coordination between create-bosh-cloudfoundry pipeline and failure-testing would be something non trivial.

combor: +1 for not having last step as a manual human signing

saliceti:
Good summary. Let's spike the signed tags quickly to have better answers.

tobwilk:
I think there is still some residual risks and considerations to document:

1. If malware is present on a users device, it could still interfere with the tag signing at the time it is carried out. The concept of requiring multiple users to jointly sign a commit tag for release helps to reduce this risk.

2. If malware is present on an individuals device, it may attempt to steal the signing key(s) for later use, even if its stored on some sort of hardware token like a yubikey. We should explore how the stealing of this can be prevented, e.g. by locking it into the token.

3. An attacker may attempt to compromise all of the individuals needed to overcome the multiple signing process. More individuals makes this harder. Who these individuals are may matter.

4. There is a risk if the individuals devices share an underlying common platform, for example being part of the same AD. If that platform itself is compromised, it may undermine all the devices being used to sign tags.

3. If an individual looses there key OR its theft is assumed, how is its redaction and replacement going to be handled. Can this be done quickly. This is a key management consideration.

dcarley:
> If malware is present on an individuals device, it may attempt to steal the signing key(s) for later use, even if its stored on some sort of hardware token like a yubikey. We should explore how the stealing of this can be prevented, e.g. by locking it into the token.

Is this the difference between generating the private key on the Yubikey vs. importing existing keys? I'm not sure whether it's possible to disable the import of existing keys to a device though, which may mean the strongest enforcement we can make is a recommendation.

docs/architecture_decision_records/ADR0XX-controlling-prod-releases.md
```diff
@@ -0,0 +1,85 @@
...
+correctly before applying it to the environment.
+
+The production concourse will check that a tag is signed by the staging
+concourse.
+
+In order to mitigate the risks of a single laptop or credential compromise
+allowing a prod deployment we have two options:
+
+- ensure that the tag is signed by two individuals (committer and reviewer) so
+  that two sets of credentials are required
```
Jonty: It isn't possible to have two people sign the same commit, a workaround would be to have two different tags for reviewer and committer, and have those signed respectively.

bleach:
> An attacker may attempt to compromise all of the individuals needed to overcome the multiple signing process. More individuals makes this harder. Who these individuals are may matter.

> There is a risk if the individuals devices share an underlying common platform, for example being part of the same AD. If that platform itself is compromised, it may undermine all the devices being used to sign tags.

Good points.

I think it's ok to implement the tag signing and verification to addres the risk that a compromised concourse or github account alone is able to push code. Once that's in place we can look at more advanced attacks.

bleach:
Thanks for the comments everyone. I'll close this for now, since everyone had a chance to comment and our spikes are in progress now.

PR #28, opened 2016-04-19
Retro action: check for conflicting work at kickoff

bleach:
In our retrospective of 8/3/2016 we discussed some occasions when two stories
played at the same time changed the same lines of code in different ways. This
caused a number of merge conflicts that needed to be resolved before finished
stories could be merged into master.

We agreed that the best way to avoid difficult and time consuming merges was to
identify at kick-off time if was likely that the story being kicked off would
cause conflicts with a story already in play.


saliceti:
👍

PR #32, opened 2016-04-22
Add an ADR around service brokers

timmow:


mtekel:
LG. I only wonder what happened to ADR003.

keymon:
LGTM.

PR #33, opened 2016-04-22
[#115931361] Add ADR for HAProxy in front of gorouter

keymon:
[#115931361 Ensure correct HSTS headers are being served](https://www.pivotaltracker.com/story/show/115931361)

What
---

We want to serve [HSTS
headers](https://en.wikipedia.org/wiki/HTTP_Strict_Transport_Security) for all
HTTPS requests to the apps domains, but it will safeguard existing users from
being MITMed over insecure connections and it will improve the user experience
when they click on a hostname that doesn't have a protocol.
(Note that without pre-loading in browsers this won't help first time users,
but that is out of context)

In this PR we add ADR explaining the context and decision of setting up HAProxy
in front of the gorouters.

Who?
---

Anyone but @keymon or @HenryTK

### DIFF
--- a/docs/architecture_decision_records/ADR008-haproxy-for-request-rewriting.md
+++ b/docs/architecture_decision_records/ADR008-haproxy-for-request-rewriting.md
@@ -1,0 +1,102 @@
+Context
+=======
+
+We want to serve [HSTS
+headers](https://en.wikipedia.org/wiki/HTTP_Strict_Transport_Security) for all
+HTTPS requests to the apps domains, but it will safeguard existing users from
+being MITMed over insecure connections and it will improve the user experience
+when they click on a hostname that doesn't have a protocol.
+(Note that without pre-loading in browsers this won't help first time users,
+but that is out of context)
+
+We want to leave open the option of able to override these headers from
+the tenant application if they wish.
+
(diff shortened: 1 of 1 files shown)


mtekel:
All fine, except it looks hastily written.

keymon:
@mtekel how can be improved?

mtekel:
Just fix up typos, formatting etc. Also `HAProxy SSL termination has better performance than gorouter.` - this looks like important advantage and might it make seem like we're doing lots of SSL terminations on gorouter. But we currently terminate on ELBs and they keepalive their connections to gorouters (which still do terminate SSL, but from ELBs only and it's keepalive - much lower load).

keymon:
Regardless if it has a big impact or not, it is a "positive consequence" but not a motivation to replace it. But I will add a comment about it. 

I do not see issues in formatting, can you elaborate?

Can you point out the typos?

keymon:
I added the comment about ELB and SSL performance.

Not sure if I fixed about typos or issues in formatting, please point them out if required.

mtekel:
Maybe this should be ADR007? Or is 007 still coming?

keymon:
Yes, somebody else is working in a ADR in a different branch.

PR #34, opened 2016-04-25
[#117605669] Add documentation about GPG signing merge commits.

alext:
This covers how we are going to GPG sign PR merges.

### DIFF
--- a/docs/guides/GPG.md
+++ b/docs/guides/GPG.md
@@ -72,6 +72,9 @@ Write the following to your shell initialisation (e.g. `~/.bashrc` or
       fi
     fi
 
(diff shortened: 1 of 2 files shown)


mtekel:
LG, will merge after https://github.com/alphagov/paas-cf/pull/223

mtekel:
As it turns out, git is not matching only committer email addess, but whole string `Your name <your@mail.com>` - if you have a typo in name, or a space behind it or anything that makes it not match `Your mame  <in@gpg.key.list>` - then gpg won't match the key and git won't sign. You'll have to use git configuration options to explicitly specify gpg ID to use for signing then.

alext:
@mtekel I've updated this to clarify that it's matching on the full name and email.

PR #40, opened 2016-05-11
[#115874555] Update team manual to include late and banned words penalties so new …

Rory80Hz:
…team members can understand these

### DIFF
--- a/docs/team/penalties.md
+++ b/docs/team/penalties.md
@@ -1,0 +1,53 @@
+# Penalties
+There are certain things which can earn you a 'mark'. Not as terrifying or drastic as the dreaded [Black Spot](https://en.wikipedia.org/wiki/Black_Spot_(Treasure_Island)) created by Robert Louis Stevenson in his novel 'Treasure Island', but nonetheless still part of our team culture.
+
+The penalty, for achieving the most marks for a sprint, is to buy treats/healthy eats for the team retrospective.
+
+The markable offences are listed below.
+
+## Late for Standup
+Scary. Turning up late for the standup? How could such a thing happen. One Mark.
+
+Missing standup, or being late, means you will miss out on updates on:
+
+* Team News (who is where today)
+* Product News (any significant developments worth knowing about)
+* Context on Work in Progress
+
+The last one is especially important, as the standup is a valuable way to crowdsource ideas on problems that people may be having with a story, and if you are not there, you can't help.
+
+## Use of Banned words
+Diversity is important to us, as is inclusivity. To help create such an atmosphere we avoid the use of certain words. Use any of the words below and you receive One Mark.
+
+
+### Gender Specific Words
+| Gender Specific Words |
+|---	|---	|---	|---	|
+| guys | gents | chap | man |
+| he | girls | &nbsp; | &nbsp; |
+
+There is a great write up on the word [guys by Jessica Evans](http://jvns.ca/blog/2013/12/27/guys-guys-guys/) that helps explain why it isn't quite gender neutral.
+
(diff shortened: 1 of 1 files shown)


docs/team/penalties.md
```diff
@@ -0,0 +1,43 @@
...
+There is a great write up on the word [guys by Jessica Evans](http://jvns.ca/blog/2013/12/27/guys-guys-guys/) that helps explain why it isn't quite gender neutral.
+
+Another great reference article [the first rule of devops club by Bridget Kromhout](http://bridgetkromhout.com/blog/2014/11/03/the-first-rule-of-devops-club/)
+
+### Assumed expertise
+| Assumed expertise |
+|---	|---	|---	|---	|
+| just | obviously | trivial | &nbsp; |
+
+Whilst obviously these words are just trivial throwaway adjectives on the face of it, the reality is they assume expertise and can act to belittle contributions to a discussion. For example, using the word obviously can be a form of [Hindsight Bias](https://en.wikipedia.org/wiki/Hindsight_bias), some people reading this may know it, but others may not. Not so obvious right?
```
combor: Hindsight Bias is interesting. I didn't know about it but as I understand it relates to "post-mortem" analysis. So saying "It obviously failed, I knew it will" is an example of it. Correct me if I'm wrong - I didn't read entire article :). In the case of "obviously" and "just" we more want to avoid underestimation of efforts. It might be a [Dunning–Kruger effect](https://en.wikipedia.org/wiki/Dunning%E2%80%93Kruger_effect) as well.

Rory80Hz: Good point! Part of the Hindsight Bias is that something may be really obvious to someone else because they have done it before or have experience with something, and assume everyone else does.

I'll add some lines about the underestimation of effort...

docs/team/penalties.md
```diff
@@ -0,0 +1,43 @@
+# Penalties
+There are certain things which can earn you a 'mark'. Not as terrifying or drastic as the dreaded [Black Spot](https://en.wikipedia.org/wiki/Black_Spot_(Treasure_Island)) created by Robert Louis Stevenson in his novel 'Treasure Island', but nonetheless still part of our team culture.
+
+The penalty, for achieving the most marks for a sprint, is to buy treats/healthy eats for the team retrospective.
+
+The markable offences are listed below.
+
+## Late for Standup
+Scary. Turning up late for the standup? How could such a thing happen. One Mark.
```
dcarley: Can we say why this is important?

docs/team/penalties.md
```diff
@@ -0,0 +1,43 @@
...
+
+## Use of Banned words
+Diversity is important to us, as is inclusivity. To help create such an atmosphere we restrict the use of certain words. Use any of the words below and you receive One Mark.
```
dcarley: Can we soften "restrict" to "avoid"?

docs/team/penalties.md
```diff
@@ -0,0 +1,43 @@
...
+
+### Gender Specific Words
+| Gender Specific Words |
+|---	|---	|---	|---	|
+| guys | gents | chap | man |
+| he | girls | &nbsp; | &nbsp; |
+
+There is a great write up on the word [guys by Jessica Evans](http://jvns.ca/blog/2013/12/27/guys-guys-guys/) that helps explain why it isn't quite gender neutral.
+
+Another great reference article [the first rule of devops club by Bridget Kromhout](http://bridgetkromhout.com/blog/2014/11/03/the-first-rule-of-devops-club/)
```
dcarley: I'd like to expand on this a bit more, but I can do that in a subsequent PR, if you don't mind?

Rory80Hz: Yes, that would be great!

PR #49, opened 2016-07-14
[#115884129] Check diego job ordering

HenryTK:
## What

Add a reminder to check Diego job ordering when upgrading CloudFoundry. We need to do this now that we are following [the job ordering in the Diego manifest](https://github.com/cloudfoundry/diego-release/blob/v0.1476.0/manifest-generation/diego.yml#L122-L424).

## How to review

Read it. Make sure it makes sense.

## Who

Anyone but @alext or me.


mtekel:
Reviewed by @benhyland , merging.

PR #57, opened 2016-08-31
Improve docs on merging PRs and forking

benhyland:
Just some refinements to be more explicit.

Especially helpful for new starters, but also it's good to document our
current practice, rather than a historical or idealistic version of it.

## How to review

`mkdocs serve` and friends as per the README.
Visit `localhost:8000` and make sure the changed page makes better sense than before.

## Who can review

Anyone but @benhyland

### DIFF
--- a/docs/team/working_practices.md
+++ b/docs/team/working_practices.md
@@ -105,8 +105,19 @@ on GitHub, so that someone else doesn't duplicate effort.
 
 # Merging Pull requests
 
-Pull requests must be merged using a GPG signed commit. This means that merges
-can't be done in the Github UI, and have to be done locally.
+Once review is complete, and all status checks have passed (Travis etc),
+a pull request can be merged.
+
+PRs to the [paas-cf][] repository should be merged using a GPG signed commit.
+This means that merges can't be done in the Github UI - they have to be done locally,
+using the `merge_pr` make task.
+
+We only use signed revisions of [paas-cf][] from our concourse pipelines. This is enforced
+by the gpg functionality in [paas-git-resource][]. We may require signing merges in other
+repositories in future.
+
+[paas-cf]: https://github.com/alphagov/paas-cf
+[paas-git-resource]: https://github.com/alphagov/paas-git-resource
 
 ## Initial setup for signing commits
 
@@ -133,8 +144,8 @@ Github can verify commits that you sign as described
 
 ## Merging a branch into master
 
-We have implemented a helper script to handle merging PR in some repos (eg
-paas-cf). For these repos PRs can be merged and signed by doing:
+We have implemented a helper script to handle merging and signing PRs for you.
+In repos where this is available (eg paas-cf) PRs can be merged and signed by running:
 
 ```sh
 make merge_pr PR=<n>
@@ -144,8 +155,7 @@ Where `n` is the PR number.
 
 ### Manually merging and signing a PR
 
-For other repos PRs will need to be manually merged and signed once review is
-complete, and all status checks have passed (Travis etc).
+PRs can also be be manually merged and signed.
 
 First check that your working directory is clean, then do the following:
 
@@ -236,7 +246,7 @@ ask a permanent member of the team to do this for you:
 [coding in the open]: https://gds.blog.gov.uk/2012/10/12/coding-in-the-open/
 
 1. Prefix the name with `paas-` so that it's easier to find. This includes
-forks of third-party repos which will need to be renamed after forking, e.g.
+forks of third-party repos which will need to be renamed after forking, eg
 `cf-release` would become `paas-cf-release`.
 1. Add the Pivotal Tracker service integration/hook. You will need to take
 the API key from the "Profile" page of your own account.
@@ -251,4 +261,6 @@ We should never pull request against the `master` branch of a forked repo.
 Doing so makes it very hard to reconcile our changes against the upstream
 repo at a later date. We should aim to submit all changes upstream. It may
 be appropriate to run against a branch of our fork for a limited period of
-time.
+time. We have decided to use a branch called `gds_master` for this, although
+some older reporitories may use something else. Feel free to change these as
+you encounter them, to improve consistency.
--- a/docs/team/working_practices.md
+++ b/docs/team/working_practices.md
@@ -125,7 +125,7 @@ Ensure you have gpg setup on your machine (see [guides/GPG](../guides/GPG) for
 details).
 
 By default, git will look for a secret key in your local keyring that exactly
-matches your configured committer name and email address (eg
+matches your configured committer name and email address (such as
 `Testy McTest <test@example.com>`). If this is not sufficient, you can
 configure the signing key that git will use:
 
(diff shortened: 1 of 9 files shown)


docs/team/working_practices.md
```diff
@@ -105,16 +105,26 @@ on GitHub, so that someone else doesn't duplicate effort.
...
+[paas-git-resource]: https://github.com/alphagov/paas-git-resource
 
 ## Initial setup for signing commits
 
 Ensure you have gpg setup on your machine (see [guides/GPG](../guides/GPG) for
 details).
 
 By default, git will look for a secret key in your local keyring that exactly
-matches your configured committer name and email address (eg
+matches your configured committer name and email address (e.g.
```
alext: This change is contrary to the GDS style-guide where we're encouraged not to use the dots.

benhyland: Citation please. https://www.gov.uk/guidance/style-guide/a-to-z-of-gov-uk-style suggests we just shouldn't use it at all. IMO I am not a content writer and if you expect this level of detail from my docs you need a copy-edit phase.

alext: Ah, I think the guidance has moved further since then, but certainly, everywhere in GOV.UK content you'll find it as "eg", and not "e.g.", which is why it was this way here. For this reason, I don't think we should be changing this.

benhyland: Do we have any tools to help us make the build fail when something like this slips through? I don't think it's an appropriate subject for PR review, unless we have an explicit copy-edit phase in our process.

Jonty: When you are editing a repo that is entirely documentation I'd argue that the PR review _is_ that phase.

benhyland: I disagree. The repo is not entirely documentation, the guidelines quoted are hard to apply case by case by non-experts and in any case optimise for a different audience.

Jonty: * Fine: This repo is _primarily_ documentation - the Makefile and yaml configuration do not change that.
* The point of review on a documentation repo is to check that the documentation is correct, and also to lightly check language and grammar. This is absolutely the right place to be pointing out anything you're aware might be wrong.
* You changed language for subjective reasons unrelated to the main purpose of the PR, and someone said "could you not, here's why".

We do not expect people to be copy editors or even be aware of the GDS standards, however when one team member spots something they know to be out of line in any context and calls it out we listen to them, especially when the change is otherwise needless.

timmow: Agreed.

I also think the benefits of adding automated testing for this sort of error will do not warrant the amount of time required to implement them.

benhyland: @timmow I strongly disagree that we should be attempting to enforce these rules at all without automation backing it up, but since that is the team decision I have fixed my original commit and added another to bring these docs in line with that particular rule in the styleguide.

To expand on my position a little:
If @alext had only said 'don't do this now, it's an unrelated change' I could understand.
But the argument seems to be that we should (manually!) apply an arbitrary, incomplete selection of rules, in a way which doesn't conform to the existing style of the repo, and in a way which in inconsistent across PRs. I don't agree that this is a good process: it doesn't improve overall quality and it is actively hostile to friction-free change.

Jonty: My point was not that we should apply an arbitrary set of rules, it was that something is pointed out in review that is known to be incorrect we should fix it immediately rather than leave it in favour of solving the generalised problem in the future.

Yes, we could automate some of this. Yes, we should have better guides and documentation. If we manage to programmatically formalise the GDS english language style guidelines I'm sure the content team would be delighted.

In the meantime, when someone points out something that is wrong we fix it and keep a note of how often it happens and generates friction - when it starts to become a serious irritation to the whole team we prioritise work to actually solve the problem. If we attempt to fix every tiny issue as soon as we encounter it we'll have a perfect development environment and workflow, but no product.

benhyland: @Jonty Perhaps the source of my irritation is that I value consistency way more than I value a particular style fix, so it seems to me that picking up on this in individual commits isn't actually improving the repo in a meaningful way.
For some changes, incremental improvements with the buy-in of the whole team to fix areas as they touch them can be worthwhile, although even then, the approach should be clearly communicated in advance - I just don't think this falls into that category.

docs/team/working_practices.md
```diff
@@ -105,8 +105,18 @@ on GitHub, so that someone else doesn't duplicate effort.
...
-can't be done in the Github UI, and have to be done locally.
+Once review is complete, and all status checks have passed (Travis etc),
+a pull request can be merged.
+
+PRs to the [paas-cf][] repository should be merged using a GPG signed commit.
+This means that merges can't be done in the Github UI, and have to be done locally.
+
+We only use signed revisions of [paas-cf][] from our concourse pipelines. This is enforced
+by the gpg functionality in [paas-git-resource][]. We might apply this restriction to other
+git resources in future.
```
mtekel: git repositories?

docs/team/working_practices.md
```diff
@@ -105,8 +105,18 @@ on GitHub, so that someone else doesn't duplicate effort.
 
 # Merging Pull requests
 
-Pull requests must be merged using a GPG signed commit. This means that merges
-can't be done in the Github UI, and have to be done locally.
+Once review is complete, and all status checks have passed (Travis etc),
+a pull request can be merged.
+
+PRs to the [paas-cf][] repository should be merged using a GPG signed commit.
+This means that merges can't be done in the Github UI, and have to be done locally.
```
mtekel: ...locally using merge_pr - as this does a bit more than merging (checks checks have passed, deletes branch)

docs/team/working_practices.md
```diff
@@ -251,4 +260,5 @@ We should never pull request against the `master` branch of a forked repo.
 Doing so makes it very hard to reconcile our changes against the upstream
 repo at a later date. We should aim to submit all changes upstream. It may
 be appropriate to run against a branch of our fork for a limited period of
-time.
+time. We tend to use a branch called `gds` or `gds_master` for this.
```
mtekel: I wouldn't say we _tend_. I think we have agreed that we want to use `gds_master` everywhere & fix where this is not case (not as a high priority thing). For all new forks, it should be `gds_master`. (so, its not a matter of preference or choice between `gds` and `gds_master`)

docs/team/working_practices.md
benhyland: I'll clarify this slightly.

docs/team/working_practices.md
benhyland: I meant to say resources, since I am referring to the pipeline concept rather than the repo, and it's the pipeline mechanism which is able to enforce the signing.
If this is unclear, any suggestions on a change?

docs/team/working_practices.md
benhyland: I'll clarify this.
Thanks for the comments!

docs/team/working_practices.md
mtekel: There are 2 git resources - one official and one we forked to add GPG support. I am not sure which other git _resources_ do you mean and why would we need to modify them to support GPG? I thought that perhaps you mean to say that we are going to use `paas-git-resource` for other repositories as well? Meaning that we'd apply the same workflow there (signing merges).

benhyland: That's exactly what I meant. I'll rephrase.

mtekel:
Thanks for fixes, merging.

PR #58, opened 2016-09-02
Make it clear where to obtain the pivotal token

Jonty:
What
====
I spent a few minutes clicking around to locate this, but the URL is static so we can link it from the team manual.

How to review it
============
Check the manual still renders and the link is correct.

Who can review it
=============
Not @jonty.


mtekel:
works. I even found out I had my own API token. I didn't know.

PR #60, opened 2016-09-13
Fix broken chrome plugin link

mtekel:
### What
We have moved chrome plugin location in paas-cf. Update the link.

### Reviewing
`mkdocs serve`, click on the link, should lead to correct destination

### Who
not me

### DIFF
--- a/docs/team/dashboard_mac_mini.md
+++ b/docs/team/dashboard_mac_mini.md
@@ -49,7 +49,7 @@ The combined dashboard is
 
 We're using a custom Chrome plugin to hide some of the Concourse page furniture
 on the Dashboard. It can be found
-[here](https://github.com/alphagov/paas-cf/tree/master/concourse/chrome_plugin).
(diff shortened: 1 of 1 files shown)


docs/team/dashboard_mac_mini.md
```diff
@@ -49,7 +49,7 @@ The combined dashboard is
 
 We're using a custom Chrome plugin to hide some of the Concourse page furniture
 on the Dashboard. It can be found
-[here](https://github.com/alphagov/paas-cf/tree/master/concourse/chrome_plugin).
+[here](https://github.com/alphagov/paas-cf/tree/master/misc/chrome_plugins).
```
alext: We should probably link to the subdirectory under this - https://github.com/alphagov/paas-cf/tree/master/misc/chrome_plugins/clean_concourse_pipeline

mtekel: fixed

PR #61, opened 2016-09-14
Add overview of our AWS network layout

benhyland:
Derived from notes I've made as I've been learning what
AWS does and how we use it.

I think this could be useful for others, especially new starters,
and is at a high enough level to not go instantly out of date.

To review:
Decide if the doc is useful.
If it is, check carefully for inaccuracies.
Also check for statements which might change soon or often - these should be avoided in docs which aren't tested or generated.

### DIFF
--- a/docs/team/networking_in_aws.md
+++ b/docs/team/networking_in_aws.md
@@ -1,0 +1,121 @@
+# Networking
+
+This page describes the network layout in a single PaaS deployment on AWS.
+
+## VPC
+
+We use a single [virtual private cloud](https://aws.amazon.com/documentation/vpc/) (VPC), with IP range `10.0.0.0/16`.
+
+Our VPC is in the `eu-west-1` region, in Ireland, and includes [subnets](http://docs.aws.amazon.com/AmazonVPC/latest/UserGuide/VPC_Subnets.html)
+in each of the `eu-west-1a`, `eu-west-1b`, and `eu-west-1c` [availability zones](http://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Concepts.RegionsAndAvailabilityZones.html) (AZs).
+
+Since we run a VPC for each of our CI, Staging and Production environments, and for each developer's environment, we sometimes run into [limits](http://docs.aws.amazon.com/AmazonVPC/latest/UserGuide/VPC_Appendix_Limits.html) on the resources we are allowed to use.
+We've raised various limits by submitting requests to Amazon.
+The result (sometimes out of date) is recorded in a [spreadsheet](https://docs.google.com/spreadsheets/d/1ZkHASsyROKiixrvpV-ecePMqAj9zPAjQENQlsOh6QBc/edit#gid=0).
+
+## Subnets
+
+We have 5 subnets in each AZ.
+
+`infra` handles outbound internet traffic.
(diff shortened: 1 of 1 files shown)


benhyland:
Discussed with @alext, this will definitely need rework if we decide to keep it - in particular, it doesn't mention the VPC Router and the diagram is horrible.

docs/team/network_topology.md
```diff
@@ -0,0 +1,96 @@
...
+
+Gateways have to appear in the routing table for a subnet for instances to use them.
+
+### Elastic IPs
+
+We allocate an [Elastic IP](http://docs.aws.amazon.com/AmazonVPC/latest/UserGuide/vpc-ip-addressing.html#vpc-eips) address (EIP) to each NAT Gateway, and to each of the bosh and concourse EC2 instances.
+
+The NAT Gateways need an EIP to function.
+
+The bosh and concourse VMs need EIPs for access during bootstrap, but not afterwards.
```
alext: I don't think this is strictly true. I think the EIP is necessary for them to be able to make outbound connections to the Internet (an instance in a public subnet needs a public IP or an EIP to be able to connect out). We use an EIP as opposed to just a public IP so that we can pre-allocate the address with terraform, and then insert it in the manifests.

docs/team/network_topology.md
```diff
@@ -0,0 +1,96 @@
...
+We don't currently disassociate the EIPs from the bosh and concourse instances after bootstrap.
+This is because we prefer our infrastructure to be immutable, and because the additional access could be useful if something goes wrong.
+
+### ELBs
+
+We allocate several [Classic Elastic Load Balancers](http://docs.aws.amazon.com/elasticloadbalancing/latest/classic/introduction.html) (ELBs) to provide load-balanced SSL access to Concourse, logs and metrics, and some CF components.
+
+An ELB has a presence in each AZ, and scales automatically with the traffic volume being balanced.
+The scaling is transparent to the external user because each ELB provides a DNS name - IP addresses of additional ELB components are added automatically when the ELB scales.
+ELBs also handle SSL termination.
```
alext: Might be worth linking to ARD007 here.

benhyland:
Learned some more, reworked a little bit, removed unhelpful diagram. Ready for review again.

docs/team/networking_in_aws.md
```diff
@@ -0,0 +1,119 @@
...
+
+The special destination `local` routes traffic to any subnets in the VPC.
+
+There is a main route table, which is default for any subnet without its own.
+Our main route table only specifies a `local` route.
+
+The route tables for the `cell`, `cf`, and `router` subnets also route internet traffic to the NAT Gateway for the AZ.
+
+The route table for the `infra` subnets route internet traffic to the Internet Gateway.
+The `infra` subnets are public, and are where the NAT Gateways are located.
```
saliceti: You could mention VMs in the infra subnet can't route through the NAT gateway because of the internet gateway in the same subnet. The internet gateway provides routing for VMs with public IPs, and the route table must target the internet gateway. So we can't have another similar route for the NAT gateway.

benhyland: Done

saliceti:
Thanks!

PR #62, opened 2016-09-14
[#129039813] Add ADR for SSL only policy and HSTS preload implementation

keymon:
[#129039813 Implement HSTS Preload list requirements](https://www.pivotaltracker.com/n/projects/1275640/stories/129039813)

What
===

We want to serve only HTTPS for all endpoints and implement HSTS preload. Include 2 ADRs to describe the decisions taken:

 * Add ADR for HSTS preload using AWS API Gateway

   Describe the decisions taken to implement HSTS preload responder running on
   AWS API Gateway and Cloud Front.

 * Add ADR for SSL only for all endpoints

   Document the decision of only serve HTTPS/SSL connections to any public
   endpoint, for deployed applications or PaaS services.

How to review?
-----------

Read it, check if it makes sense

Who?
---

Anyone but @keymon

### DIFF
--- a/docs/architecture_decision_records/ADR443-ssl-only-for-applications-and-cf-endpoints.md
+++ b/docs/architecture_decision_records/ADR443-ssl-only-for-applications-and-cf-endpoints.md
@@ -1,0 +1,52 @@
+Context
+=======
+
+It is expected for the government websites to be secure and keep the user
+interactions private. Because that we want to enforce all communications to
+any application and to the platform endpoints to use only and always HTTPS.
+
+When a user inputs a website name without specifying the
+protocol in the URL, most browsers will try first the HTTP protocol by default.
+Even if the server always redirect HTTP to HTTPS, an initial
+unprotected request including user information will be transferred
+in clear: full URL with domain, parameter, cookies or browser meta-information.
+
+[HTTP Strict Transport Security](https://en.wikipedia.org/wiki/HTTP_Strict_Transport_Security)
+mitigates this issue by instructing modern browsers that support it to
+always connect using HTTPS.
+
+There is still a potential initial unprotected HTTP request that might happen
+before retrieve the HSTS headers or after the specified HSTS `max-age`.
+To solve this issue, the root domain can be added to
+[HSTS preload list](https://hstspreload.appspot.com/) which will be used by most
+common browsers.
+
+Currently the only way to avoid any clear text HTTP interaction is closing or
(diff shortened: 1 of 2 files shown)


docs/architecture_decision_records/ADR443-ssl-only-for-applications-and-cf-endpoints.md
```diff
@@ -0,0 +1,52 @@
...
+
+It is expected for the government websites to be secure and keep the user
+interactions private. Because that we want to enforce all communications to
+any application and to the platform endpoints to use only and always HTTPS.
+
+When a user inputs a website name without specifying the
+protocol in the URL, most browsers will try first the HTTP protocol by default.
+Even if the server always redirect HTTP to HTTPS, an initial
+unprotected request including user information will be transferred
+in clear: full URL with domain, parameter, cookies or browser meta-information.
```
Jonty: This isn't quite true: Cookies with the "secure" flag set will only be served over HTTPS, and that flag is set by default by most frameworks when serving secure requests. I don't think that needs to be made clear in this, but worth mentioning.

Jonty:
S'good.

docs/architecture_decision_records/ADRXXX-hsts-preload-using-api-gateway.md
```diff
@@ -0,0 +1,48 @@
...
+
+To add our domains to [HSTS preload lists](https://hstspreload.appspot.com/), there are these requirements:
+
+ 1. Serve a valid certificate.
+ 2. Redirect from HTTP to HTTPS on the same host.
+ 3. Serve all subdomains over HTTPS (actually checks for `www.domain.com`)
+ 4. Serve an HSTS header on the base domain for HTTPS requests:
+
+We cannot use the Cloud Foundry app endpoint, which already [serves the
+right HSTS Security header with HAProxy](ADR008-haproxy-for-request-rewriting.md)
```
HenryTK: It serves an HSTS header but it is not 'right' in relation to HSTS preload requirements. [Our manifest](https://github.com/alphagov/paas-cf/blob/master/manifests/cf-manifest/manifest/020-cf-properties.yml#L131) shows only the `max-age` directive is added, whereas the preload list requires that the `includeSubDomains` and `preload` directives are also set.

keymon:
Ups, I forgot to mention that this should not be merged before renaming the files! also, there are some comments to address.

keymon:
I will raise a new PR to address the issues.

PR #64, opened 2016-09-29
Add guide for common support tasks

Jonty:
We've not been documenting how to carry out some common support tasks, so this introduces a new guide for them. For now it just covers changing a users password.


benhyland:
lgtm

PR #65, opened 2016-10-03
Add a easy command to generate the PR wall board url

keymon:
This will make it easier to generate the dashboard URL with some exported variables.

I suggest that the users store the credentials using pass:

```
pass insert -m github_creds <<EOF
export GITHUB_API_TOKEN=<your github access token>
export GITHUB_GDS_API_TOKEN=<your github.gds access token>
EOF
```

And then load then with:

```
eval $(pass github_creds)
```


benhyland:
good advice

PR #68, opened 2016-10-11
Document the new paas-alerts-{dev,ci,staging,prod} groups

richardc:
In playing
https://www.pivotaltracker.com/n/projects/1275640/stories/131355243 we
added new groups to send alerts to.  This is where they are.


benhyland:
Confirmed these groups exist.

PR #70, opened 2016-11-03
[#132736941] Update mac-mini docs to mention smashing

benhyland:
We've introduced a new dashboard, and also decided on using
`frame-splits` if we need to display split-screen dashboards.

## Review

Check the docs make sense.
Review with https://github.com/alphagov/paas-cf/pull/587

### DIFF
--- a/docs/team/dashboard_mac_mini.md
+++ b/docs/team/dashboard_mac_mini.md
@@ -1,12 +1,18 @@
 # Dashboard Mac Mini
 
-There's a Mac Mini attached to big screens in our team area. It shows the
+There's a Mac Mini attached to big screens in our team area. It can show the
 following:
 
 * Outstanding pull requests
 * Grafana dashboard
 * Concourse pipelines
 * Pingdom public status page
+* Overview of Datadog triggered monitors
+
+Some combination of the above might be shown according to the whims of the team.
+If we want to display multiple different views on a single screen, we use
+[frame-splits](https://github.com/dsingleton/frame-splits), downloaded locally to the mac-mini to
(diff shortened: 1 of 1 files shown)


docs/team/dashboard_mac_mini.md
```diff
@@ -1,12 +1,18 @@
...
 * Outstanding pull requests
 * Grafana dashboard
 * Concourse pipelines
 * Pingdom public status page
+* Overview of Datadog triggered monitors
+
+Some combination of the above might be shown according to the whims of the team.
+If we want to display multiple different views on a single screen, we use
+[frame-splits](https://github.com/dsingleton/frame-splits), downloaded locally to the mac-mini to
+avoid worrying about sending sensitive urls to the outside world.
```
paroxp: I'll change that.

benhyland: good catch, I asked the reviewer of alphagov/paas-cf#587 to handle this

paroxp: I've noticed, Datadog allows to create iframe boxes on the ScreenBoards.

Would it not be a better approach, if we can programatically add this to our terraform code, rather than an individual setting up local application to do the same thing?

screen shot 2016-11-04 at 10 52 28

paroxp: We don't display the screenboards, no. 

But, there is an essential desire, to move over to datadog completely. Wouldn't that make sense, not to use `frame-splits` or the hosted version by `github.io`?

benhyland: If you believe that is true I suggest you raise a ticket. I think it is out of scope for this story.

PR #71, opened 2016-11-08
Add link to the paas support manual

benhyland:
Currently it lives in google docs.

We want to eventually move it to the team manual, we're tracking that in
https://www.pivotaltracker.com/n/projects/1275640/stories/127987443

We'll need to make sure existing editors are happy with git before we move it,
so just link for the moment.


mtekel:
link works correctly, merging

PR #72, opened 2016-11-16
Update mac mini advice to mention the new machine

benhyland:
It's on a different network.
Also give more specific instructions on how to use screen sharing.


Rory80Hz:
LGTM

PR #73, opened 2016-11-17
Some tweaks to signoff process

timmow:
- Mention signoff involves checking story is in production
- Remove Tech Lead and Tech Arch commenting on story as this does not
happen
- Mention we may delegate at standup if the board is blocked


Rory80Hz:
LGTM

PR #76, opened 2016-11-30
Work around broken search integration

benhyland:
MkDocs and Readthedocs don't currently agree on how to do search.
This means that search has poor UX on the publically deployed docs
on readthedocs.io.

Here's a workaround suggested in github comments that we could use
to fix things for us while we wait for the two projects to reconcile.

See:
https://github.com/rtfd/readthedocs.org/pull/2289
https://github.com/rtfd/readthedocs.org/issues/1088

See also https://github.com/alphagov/paas-developer-docs/pull/32

Should we use this fix?

### DIFF
--- a/docs/js/fix_search.js
+++ b/docs/js/fix_search.js
@@ -1,0 +1,27 @@
+// See various broken search issues, but especially:
+// https://github.com/rtfd/readthedocs.org/pull/2289
+// https://github.com/rtfd/readthedocs.org/issues/1088
+
+function fixSearch() {
+  var target = document.getElementById('rtd-search-form');
+  var config = {attributes: true, childList: true};
+
+  var observer = new MutationObserver(function(mutations) {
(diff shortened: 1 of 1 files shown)


benh-gds:
Deploying this doesn't seem to have fixed search - searches go to the readthedocs search engine, but never find matches, even for words that are definitely in the docs. Looking at the description of the fix, it seems like it's supposed to use a "builtin search.html page" instead of https://readthedocs.org/search/, but that doesn't seem to be happening.

image

Any ideas?

benhyland:
@benh-gds it's not been deployed correctly, different behaviour from locally. We'll take a look.

benh-gds:
OK. Thanks for spotting that the search was not working as intended - I just thought it was bad. Please bear in mind that we'll be ditching mkdocs very soon so if you can't get the fix working, I'd just roll it back.

PR #80, opened 2016-12-16
Document manual process for emailing tenants

benhyland:
We've been doing this for a while, it's time we wrote it down.

This process will hopefully be superseded fairly soon, but now
we have a home for docs as the process is iterated.

### DIFF
--- a/docs/team/notifying_tenants.md
+++ b/docs/team/notifying_tenants.md
@@ -43,6 +43,8 @@ done | grep '@' | sort | uniq
 
 ## Send the email
 
+**IMPORTANT: Remember to put the recipient list in `bcc:`!**
+
 We send the email from any team member's `gov.uk` GMail account.
 
 We use support email address as a non-alias to send from. To configure that for your account, follow these instructions:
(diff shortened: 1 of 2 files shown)


docs/team/notifying_tenants.md
```diff
@@ -0,0 +1,52 @@
...
+  cf target -o $org >/dev/null;
+  for space in `cf spaces 2>&1 | tail -n+4`; do
+    cf space-users $org $space;
+  done;
+done | grep '@' | sort | uniq
+```
+
+## Send the email
+
+We send the email from any team member's `gov.uk` GMail account.
```
mtekel: +We use support email address as a non-alias to send from. To configure that for your account, follow two links below

timmow:
Can we link to this from [this page](https://government-paas-team-manual.readthedocs.io/en/latest/guides/common_support_tasks/)?

docs/team/notifying_tenants.md
benhyland: done

benhyland:
@timmow yep

mtekel:
I would add a note to use bcc for the list of recipients, so that tenants can't see who else has account as well...

benhyland:
@mtekel if you look closely, you should find this note already exists. Please feel free to add a commit making it more obvious!

dcarley:
> @mtekel if you look closely, you should find this note already exists. Please feel free to add a commit making it more obvious!

I overlooked it the first time too. Not sure if that's because I was viewing the raw instead of rendered Markdown. It wouldn't hurt to put it as the first paragraph in the section though.

mtekel:
hm, thanks for notifying, but I had to use search to find it :)

benhyland:
@mtekel @dcarley Please, if you can do better, don't hesitate

mtekel:
Before I merge, I want to get OK for the last change I did...

benhyland:
@mtekel lgtm thanks!

PR #81, opened 2016-12-20
[#136187337] Mention stack release in CF upgrade

saliceti:
# What

Story: [Upgrade cflinuxfs2 stack](https://www.pivotaltracker.com/story/show/136187337)

The stack was recently upgraded because of CVEs. We should keep track of the stack as part of our standard CF upgrade process.

# How to review
Does it make sense?

# Who can review
Not me


mtekel:
I'd add one more bullet point before all 'Check the <release>' points, which would say st. like:
* For each standalone release we use, that is also included in the CF version release we are upgrading to, review if the version included in the CF release satisfies our requirements for the standalone release. If so, remove the standalone release and use one included in CF release.

mtekel:
Thanks, looks good, merging.

PR #83, opened 2017-01-05
Update dashboard Mac Mini documentation

henrytk:
## What

There is no Pivotal story for this as it is a quick and dirty support task.

Remove some jargon, update to say we don't display Pingdom status page
anymore (but it is still available), and add a section on the use of
Automator to open the browser upon login.

## How to review

Read through and check it makes sense.

## Who

Anyone but me.


dcarley:
Explanations about why the Pingdom page isn't on the monitors and why the Automator script isn't committed are 👍

PR #84, opened 2017-01-06
[#136558969] Add IPSec debugging guide

paroxp:
## What

We're about to implement the IPSec encryption, between some of our nodes. It may so happen, that one will be in need of debugging the IPSec connectivity. Wrong time in the wrong place, I suppose.

In order to make their task simpler, we'd like to document some of the techniques we've used while implementing the IPSec.

## How to review

Read through the docs and see if these make any sense...

## Who can review

Not @mtekel or @paroxp

### DIFF
--- a/docs/guides/IPSec_debugging.md
+++ b/docs/guides/IPSec_debugging.md
@@ -1,0 +1,78 @@
+## IPSec debugging
+
+
+There may be a need, to debug the IPSec on our machines. In order to do that, you may want to log into `bosh-cli` and then `bosh ssh {router|cell}`.
+
+### Fiddle with the logging level
+
+Once logged into the VM, go ahead and edit the following file:
+
+``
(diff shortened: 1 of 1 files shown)


Jonty:
Fixed some minor typos, other than that A+ would debug ipsec again.

combor:
It might be worth to add `tcpdump -i eth0 -n -s 0 -vv \(port 500 or port 4500 or proto 50\)` command to the guide. It allows you to capture IPSEC only. I found lit quite useful.

combor:
Oh, my...too late.

mtekel:
If you open a small PR to do that I can merge quickly as I have used that command, so I have reviewed it works...

combor:
@mtekel here you go https://github.com/alphagov/paas-team-manual/pull/87

PR #85, opened 2017-01-10
Remove DEPLOY_ENV from sts token filename.

alext:
Following an update to the script, this is no longer included in the
filename.

This should be reviewed along with https://github.com/alphagov/paas-cf/pull/709


dcarley:
👍

PR #86, opened 2017-01-10
[#136783629] Update ADR009

saliceti:
# What
Story: [How do encrypt traffic between the elb and gorouter using haproxy](https://www.pivotaltracker.com/story/show/136783629)

https://github.com/alphagov/paas-cf/pull/713 and
https://github.com/alphagov/paas-haproxy-release/pull/8 encrypt the traffic between the router ELB and HAProxy.
The ADR is updated to reflect the change.

# How to review
Does it make sense?

# Who can review
Not @keymon or me


combor:
LGTM. Merging.

PR #90, opened 2017-01-17
[#136783629] Rename ADR0015 to ADR016

dcarley:
Because all of the other ADRs have 3 digit numbers and ADR015 is now taken.
Spotted by Michal. Also update the one reference to this, in a superseded
ADR.


mtekel:
All fine. One small improvement though: it would be nice if ADR016 reference in ADR009 was a clickable link

dcarley:
> All fine. One small improvement though: it would be nice if ADR016 reference in ADR009 was a clickable link

Good idea. I've pushed a commit to do that. You can test it with `mkdocs serve`.

mtekel:
thanks, that works, merging

PR #91, opened 2017-01-27
Updates from retro

bleach:
## What

We agreed on 13th December that we should document that:

- Part of approval is ensuring that follow-up work is scheduled
- Definition of done should include keeping users informed about changes

## Who can review 

Anyone but @bleach


mtekel:
LG

PR #93, opened 2017-02-10
How to deal with uncontactable org managers

bleach:
## What

In https://gaap.deskpro.com/agent/#app.tickets,t:3627 we discovered that
someone who was the sole org manager for a trial organisation had left that
organisation.

Documenting what we decided to do in this and future cases.

## Who can review

Anyone who does support, apart from me

## How to review 

- Is it clear what to do?
- Does @urmishah or Jess agree with the process?


alext:
Change looks good to me. I'll leave Jess or @urmishah to have the final say before this is merged.

urmishah:
Looks good to me.

PR #99, opened 2017-03-01
[#135954607] Add support section

combor:
## What

This is a placeholder for new support section that will in the future
contain info related to our support processes. Now it has only two
sub-sections pointing to google docs.

## How to test

Sanity check

## Who

Not me


mtekel:
Links work and there are no typos, merging.

PR #100, opened 2017-03-08
[#142066597] ADR for accessing user provided services in AWS

bleach:
## What

Jotted down an overview of what I think are the three main options for helping
a currently onboarding service access their hosted services.

This is to share with that team and also get an initial review from tech leads

## How to review

* Read, and verify it makes sense, and matches what we have decided.
* Optionally: squash down some of the commits before merging.

## Who can review

@dcarley @henrytk @timmow

### DIFF
--- a/docs/architecture_decision_records/ADR019-accessing-user-provided-services.md
+++ b/docs/architecture_decision_records/ADR019-accessing-user-provided-services.md
@@ -1,0 +1,81 @@
+
+Context
+=======
+
+A service that's onboarding at the moment depends on MySQL, memcached and
+elasticsearch that we don't yet offer, but are likely to in future, and has
+chosen to host them on AWS using RDS MySQL, Elasticache memcached and hosted
+Elasticache.
+
+We wanted to ensure that only the app instances belonging to that service could
+connect to those services in order to avoid other services being able to access or modify their data.
+
+- MySQL requires authentication and allows TLS
+- Elasticache memcached does not provide authentication or TLS
+- AWS Hosted elasticsearch [appears to provide the ability to sign
+  elasticsearch requests using an AWS access
+key](https://aws.amazon.com/blogs/security/how-to-control-access-to-your-amazon-elasticsearch-service-domain/).
+Requiring a custom elasticsearch client.
+
+The options for authentication are:
+
+Via Internet, using network access controls
+-------------------------------------------
+
+This is described for
+[elasticache](http://docs.aws.amazon.com/AmazonElastiCache/latest/UserGuide/Access.Outside.html)
+
+How we would set it up:
+
+- Tenant binds services to public IP addresses, restricts access to PaaS egress IPs.
+- PaaS team changes default [application security
+  group](https://docs.cloudfoundry.org/adminguide/app-sec-groups.html) set to
+ forbid access to those IP addresses.
+- PaaS team creates new application security group allowing access to those IP addresses.
+- PaaS team binds application security group to space(s) belonging to tenants
+
+Risks/costs:
+
+- Access via Internet IP is via NAT instance, introducing a likely single point
+  of failure and additional running costs to tenants.
+- Traffic is in plaintext over the Internet for some services.
+- PaaS team need to do work every time a tenant adds/changes service IP addresses.
+- If they forget to request this, other PaaS tenants have access to their service IPs.
+- PaaS team need to do work in every occasion a space needs access to the peered VPC.
+
+
+Using private address space (VPC peering)
+-----------------------------------------
+
+How we would set it up:
+
+- Tenant raises support request to begin setup process
+- PaaS team responds with a unique IP allocation eg. 172.16.0.0/24 for tenant to use
+- Tenant creates VPC using that address space
+- Tenant creates AWS security group(s) restricting access from PaaS VPC to expected services
+- Tenant sends VPC peering request
+- PaaS team creates new application security group allowing access to the VPC IP allocation
+- PaaS team binds application security group to space(s) belonging to tenants
+- PaaS team accepts VPC peering request
+
+Risks/costs:
+
+- Introduces a new network security boundary between VPCs; a risk of
+  accidentally introducing security group rules that allow more access from the
+  peered VPC than intended.
+- PaaS team need to do work in every occasion a space needs access to the peered VPC.
+
+Although this specific example uses VPC peering because the tenant in question
+uses AWS, we could use the same principle (us assigning IP address space and
+changing application security groups) to a VPN or some other network overlay
+technology to allow us to connect to things other than VPCs.
+
+Decision
+========
+
+TBC
+
+Consequences
+============
+
+TBC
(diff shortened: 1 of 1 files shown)


docs/architecture_decision_records/ADR019-accessing-user-provided-services.md
```diff
@@ -0,0 +1,81 @@
...
+- PaaS team binds application security group to space(s) belonging to tenants
+
+Risks/costs:
+
+- Access via Internet IP is via NAT instance, introducing a likely single point
+  of failure and additional running costs to tenants.
+- Traffic is in plaintext over the Internet for some services.
+- PaaS team need to do work every time a tenant adds/changes service IP addresses.
+- If they forget to request this, other PaaS tenants have access to their service IPs.
+- PaaS team need to do work in every occasion a space needs access to the peered VPC.
```
dcarley: Maybe use a different phrase than "peered VPC" because it makes me think of the option below.

docs/architecture_decision_records/ADR019-accessing-user-provided-services.md
```diff
@@ -0,0 +1,81 @@
...
+
+Although this specific example uses VPC peering because the tenant in question
+uses AWS, we could use the same principle (us assigning IP address space and
+changng application security grouyps) to a VPN or some other network overlay
+technology to allow us to connect to things other than VPCs.
+
+Decision
+========
+
+TBC
```
dcarley: What are the next steps? Spike whether a solution works in practice? Update this PR when we've made a decision?

docs/architecture_decision_records/ADR019-accessing-user-provided-services.md
```diff
@@ -0,0 +1,81 @@
...
+- Tenant binds services to public IP addresses, restricts access to PaaS egress IPs.
+- PaaS team changes default [application security
+  group](https://docs.cloudfoundry.org/adminguide/app-sec-groups.html) set to
+ forbid access to those IP addresses.
+- PaaS team creates new application security group allowing access to those IP addresses.
+- PaaS team binds application security group to space(s) belonging to tenants
+
+Risks/costs:
+
+- Access via Internet IP is via NAT instance, introducing a likely single point
```
dcarley: I'm still a bit fuzzy about this. Why does it need to be behind NAT? Why could it not be a NAT Gateway (rather than instance)?

docs/architecture_decision_records/ADR019-accessing-user-provided-services.md
mtekel: It would be nice to have a "vpn" broker, which would automate these steps. When you create it, it gives you IP. When you bind it to app, it adjusts security group for the app to be able to connect.

The catch/pain is managing IP ranges. E.g. when someone would want to connect their existing service in their existing IP range and that range overlaps with some other. That covers some "hybrid" deployments where DB is in one place and access to it is from PaaS and elsewhere, and the DB needs to stay where it is with the IPs it has. The solution broker would need to implement then would be more complicated and possibly include NAT or do some other custom routing.

docs/architecture_decision_records/ADR019-accessing-user-provided-services.md
```diff
@@ -0,0 +1,81 @@
...
+elasticsearch that we don't yet offer, but are likely to in future, and has
+chosen to host them on AWS using RDS MySQL, Elasticache memcached and hosted
+Elasticache.
+
+We wanted to ensure that only the app instances belonging to that service could
+connect to those services in order to avoid other services being able to access or modify their data.
+
+- MySQL requires authentication and allows TLS
+- Elasticache memcached does not provide authentication or TLS
+- AWS Hosted elasticsearch [appears to provide the ability to sign
```
lgarvey: We have put together some sample PHP code that performs Sig v4 signing and makes a successful request to the ES endpoint.  As ES is not VPC specific this should be unaffected by the decision to use VPC peering or not. ES also provides a HTTPS and HTTP endpoint, but I have not yet found a way to force HTTPS.

docs/architecture_decision_records/ADR019-accessing-user-provided-services.md
lgarvey: @dcarley the requirements are listed in the "accessing elasticache outside of AWS" article: http://docs.aws.amazon.com/AmazonElastiCache/latest/UserGuide/Access.Outside.html

It's probably because custom iptables routes are required.  

However, it's worth noting the "Limitations" section, which stipulates: "This approach should be used for testing and development purposes only."

I don't see a clean way of using elasticache if the VPC's aren't peered.

docs/architecture_decision_records/ADR019-accessing-user-provided-services.md
bleach: That was a copy/paste. I have changed it now.

docs/architecture_decision_records/ADR019-accessing-user-provided-services.md
bleach: @dcarley my preferred next step is to spike the private IP / VPC peering option. Which seems like the most sensible approach for us now.

bleach: @mtekel For now, we only have this one concrete use case, so I think we should try it out before brokering / automating. We're talking to other teams at the moment who might need *something*, but I don't know for certain that we need a VPN for all of them yet.

I am not overjoyed about maintaining a IP address registry at all, but it seems better than the alternatives at the moment.

docs/architecture_decision_records/ADR019-accessing-user-provided-services.md
```diff
@@ -0,0 +1,86 @@
...
+- Introduces a new network security boundary between VPCs; a risk of
+  accidentally introducing security group rules that allow more access from the
+  peered VPC than intended.
+- PaaS team need to do work (assigning application security groups) on every
+  occasion a space needs to gain or lose
+  access to the peered VPC.
+
+Although this specific example uses VPC peering because the tenant in question
+uses AWS, we could use the same principle (us assigning IP address space and
+changng application security grouyps) to a VPN or some other network overlay
```
mtekel: typo grouyps

alext:
I've updated this with details of the decision etc.

It's now ready for final review and merging.

docs/architecture_decision_records/ADR019-accessing-user-provided-services.md
```diff
@@ -0,0 +1,89 @@
...
+========
+
+We will offer VPC peering to tenants in specific cases where it is appropriate.
+
+Consequences
+============
+
+Where we don't presently offer a specific backing-service, tenants have an
+option to provision their own service and access it without having to expose it
+to the Internet.
```
dcarley: @alext Should we add something here about the manual steps we need to perform? Perhaps link to https://github.com/alphagov/paas-cf/blob/master/doc/vpc_peering.md ?

alext: Good shout. I've added a link.

PR #101, opened 2017-03-09
[#140553003] Test rotating credentials when doing CF upgrades

dcarley:
## What

New releases can introduce new passwords or different uses of existing
passwords. It's possible that these may not be easy to rotate without
causing downtime. We'd like to know if this is the case before we next need
to use the credential rotation in anger (e.g. leaked secrets).

This was originally part of a follow-up story but I've pulled it out after a
discussion in knowledge share because we're likely to do one or more CF
upgrades before we play the follow-up to rotate more credentials.

## How to review

Confirm that:
- new text makes sense
- is in an appropriate place in the guide
- is a sensible thing for us to do.
- the link to another page in the team manual works

I'll remove the documentation part from the story when merged.

## Who can review

Not @dcarley


mtekel:
LG and link works. I have also updated #140553003 that the documentation is done.

PR #102, opened 2017-03-16
[#139179971] Add a note to notify tenants after upgrade

keymon:
What?
----

We want to add some notes about how to notify tenants after a CF upgrade.

Include reference to the generate_buildpack_release_notes.sh script added in https://github.com/alphagov/paas-cf/pull/829

How to review?
---------

Read it, check that makes sense :)

Who?
----

Anyone but @keymon or @csaliceti


mtekel:
indeed

henrytk:
We already have a dedicated page for notifying tenants: https://government-paas-team-manual.readthedocs.io/en/latest/team/notifying_tenants/

Should we be adding to that and linking to it instead?

dcarley:
> Should we be adding to that and linking to it instead?

🙌  

The process is described there and I think we could add a new entry under templates to capture this information.

@keymon could you raise another PR to update?

PR #103, opened 2017-03-16
Update tech docs deployment guide

bleach:
## What

Update the docs to make it clear that the deployment should be automatic and also that we've finished our migration now.

## How to review 

Read the words what I wrote and confirm they can be parsed by humans & agree with reality.

## Who can review

Anyone but me. Possibly @Jonty would be interested?


mtekel:
Looks good, words and complete sentences make sense and links point to correct places, merging.

PR #104, opened 2017-03-20
[#136027405] update team manual with google login workaround details

bleach:
## What

A workaround will be required to login to admin accounts. Update the team manual to enable people to find the workaround to login easily.

## How to review

Ensure the meaning is clear. I have tested the instructions myself in the course of reviewing alphagov/paas-cf#830

If you like you may delay merging this change until alphagov/paas-cf#830 is merged, as the instructions won't work until that's been rolled out. However it seems unlikely that anyone will read them before we do roll it out anyway.

## Who can review

A team member who is not me

### DIFF
--- a/docs/guides/cf_admin_users.md
+++ b/docs/guides/cf_admin_users.md
@@ -5,6 +5,18 @@ able to attribute actions to individuals.
 
 In order to ensure two factor authentication is used, we use our google apps accounts to authenticate:
 
-`cf login --sso`
+`cf login -a api.SYSTEM_DOMAIN --sso`
 
-At present, a bug in UAA means that a workaround is necessary. This is [documented here](https://github.com/alphagov/paas-cf/pull/830#issue-214424527)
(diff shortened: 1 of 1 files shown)


docs/guides/cf_admin_users.md
```diff
@@ -0,0 +1,10 @@
+
+Everyone in the team who is authorised for prod access gets their own admin
+account in CI, staging and prod to avoid needing to share credentials and to be
+able to attribute actions to individuals.
+
+In order to ensure two factor authentication is used, we use our google apps accounts to authenticate:
+
+`cf login --sso`
+
+At present, a bug in UAA means that a workaround is necessary. This is [documented here](https://github.com/alphagov/paas-cf/pull/830#issue-214424527)
```
alext: I think it would be worth including the workaround details here instead of linking to the PR.

bleach: I've added the workaround details here now, so it should be easier to read

dcarley:
Followed this while reviewing alphagov/paas-cf#851 👍

PR #105, opened 2017-03-20
[#139179971] Move upgrade notification to the relevant page

saliceti:
# What

Story: [Upgrade CF to >= v252](https://www.pivotaltracker.com/story/show/139179971)

The CF upgrade notification part was added earlier to the upgrade section. It was decided it makes more sense to move it to the notification page.

# How to review

Does is make sense?

# Who can review
Not me


mtekel:
links work, text makes sense. merging.

PR #106, opened 2017-03-27
[#137700363] Describe manual setup for emergency support

saliceti:
# What

Story: [Set up emergency email for tenants to use when alerting us OOH](https://www.pivotaltracker.com/story/show/137700363)

* Pageduty emergency service setup
* New "Deskpro" section. Only the setup for emergency is documented. The basic setup should be added later on

# How to review

* Does it make sense?
* Do links work?
* Is the real setup the same as described?

# Who can review
Not me

### DIFF
--- a/docs/team/deskpro.md
+++ b/docs/team/deskpro.md
@@ -1,0 +1,31 @@
+[Deskpro](https://www.deskpro.com/) is the GaaP support tool. Each team customises it for its needs.
+
+You need to be admin to change the settings.
+
+*Basic manual setup should be documented here*
+
+## Emergencies
+
+## Email account
+
+A `@gaap.deskpro.com` [email account](https://gaap.deskpro.com/admin/admin-interface#/tickets/ticket_accounts) is set up for emergencies. We don't currently use an `@digital.cabinet-office.gov.uk` address because we aren't confident that emails won't get caught by moderation and spam filters.
+
+* The priority is set to `P1`.
+* It has an action `Send email`:
+  * Send to the emergency Pagerduty address
+  * Based on template `gov-uk-paas-emergency.html`
+* Set Email Account: normal support email address (otherwise the conversation continues on the emergency email address)
+
+## Custom email template
+
+The [custom email template](https://gaap.deskpro.com/admin/admin-interface#/tickets/email_templates/custom) is called `gov-uk-paas-emergency.html`.
+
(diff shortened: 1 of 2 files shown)


docs/team/pagerduty.md
```diff
@@ -38,6 +38,11 @@ The team email notification is added at the end as well to cover any potential g
 We integrate with Pingdom and Datadog. For each integration we create one *in hours* service that uses the in hours only escalation policy, and one *24x7* service that uses the 24x7 escalation policy.
 Incidents are not auto resolved. Only high-urgency notification rules are used.
 
+## Emergency
+
+A special service called *Emergency email* is dedicated to emergencies and triggered by email from Deskpro. The incoming emails are filtered by regular expression.
```
dcarley: Should this section be combined with the "Services" section, since it's another service?

dcarley: Could we describe the purpose of this better than "emergency"? Something about a mechanism for tenants to urgently tell us that something is wrong?

docs/team/deskpro.md
```diff
@@ -0,0 +1,24 @@
...
+You need to be admin to change the settings.
+
+*Basic manual setup should be documented here*
+
+## Emergencies
+
+## Email account
+
+A `@gaap.deskpro.com` [email account](https://gaap.deskpro.com/admin/admin-interface#/tickets/ticket_accounts) is set up for emergencies. A `@digital.cabinet-office.gov.uk`
+may be used instead in the future.
```
dcarley: Rather than saying what might change in the future:

> A `@digital.cabinet-office.gov.uk` may be used instead in the future."

It might be preferable to say why we didn't previously chose to do a thing:

> We don't currently use an `@digital.cabinet-office.gov.uk` address because we aren't confident that emails won't get caught by moderation and spam filters.

docs/team/deskpro.md
```diff
@@ -0,0 +1,24 @@
...
+
+* The priority is set to `P1`.
+* It has an action `Send email`:
+  * Send to the emergency Pagerduty address
+  * Based on template `gov-uk-paas-emergency.html`
+* Set Email Account: normal support email address (otherwise the conversation continues on the emergency email address)
+
+## Custom email template
+
+The [custom email template](https://gaap.deskpro.com/admin/admin-interface#/tickets/email_templates/custom) is called `gov-uk-paas-emergency.html`.
```
dcarley: Should we say something about the subject being preserved because it's the title of the alert in PagerDuty?

docs/team/pagerduty.md
saliceti: We could but as this doc is public I would rather not give hints of how to bypass the system

dcarley: From offline: probably not a risk because they'd still need to know the from and to addresses.

saliceti:
I think I've addressed the comments

dcarley:
Looks good, thanks for the updates!

PR #107, opened 2017-03-30
[#134579301] Add support security section

saliceti:
# What

Story: [Create alerts for anomalous UAA behaviour](https://www.pivotaltracker.com/story/show/134579301)

Help to diagnose brute force attacks by looking at UAA logs.

# How to review
Does it make sense?

# Who can review
Not me

### DIFF
--- a/docs/support/responding_to_alerts.md
+++ b/docs/support/responding_to_alerts.md
@@ -1,0 +1,18 @@
+Tips for responding to alerts when you are on support.
+
+## Failed logins
+
+In case of security incident, for example when alerted of a brute-force attack by datadog monitors,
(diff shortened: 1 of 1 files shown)


dcarley:
Would you mind re-purposing this into a general "alerts" section that we can add to for other kinds of alerts? Specifically:

- rename the page to "support/alerts" or "support/handling alerts" or "support/responding to alerts"
- put the content about UAA brute force under a header like "failed UAA logins" and link to that specific URL fragment from the alert

dcarley:
Content looks good, btw 👍

saliceti:
Done!

PR #109, opened 2017-04-19
[#143024609] Document the way of finding user activity

paroxp:
## What

We'd like to show we can investigate the activity of a user whose credentials have been compromised.

## How to review

- Read through the content
- Apply instructions and attempt perform little investigation

## Who can review

Not @paroxp

### DIFF
--- a/docs/support/finding_activity.md
+++ b/docs/support/finding_activity.md
@@ -1,0 +1,96 @@
+# Finding the activity for a given user
+
+In case of suspicion of compromised CF details, we can use Kibana to
+obtain some information about the API use.
+
+## Finding user GUID
+
+You may need to obtain this, to move the investigation forward. It's
+usually useful, to have a starting point in the form of organisation.
+Possessing that name, use this command, to obtain users from the
+organisation:
+
+```
+$ cf curl "/v2/users?q=organization_guid:$(cf org ORG_NAME --guid)"
+```
+
+**Note:** It may be a big chunk of JSON, which you may want to export
+into a file.
+
+Each user as an entity, will have metadata enclosed. This metadata will
+have users GUID.
+
+### Finding deleted user GUID
+
+There may be a need to find a GUID for already deleted user. There is a
+way to do that, but it's a little inconvenient...
+
+If you go to Kibana, you can use the `"DELETE \"/v2/users"` phrase in
+your search. This will print out the logs that were taken on user
+removal. This does not however, give you an indication of who the user
+was.
+
+## Kibana: Finding API usage
+
+We can make the investigation easier for ourselves by considering the
+following criteria:
+
+- Filter by user by typing `"for user: {GUID}"` in a search box
+- *Optionally:* Scope to job: `api` - From the sidebar on the left,
+  select `@source.job` and add `api` by pressing little magnifying glass
+  with a plus sign.
+- `@message` should consist of:
+	- Detailed endpoint - You may filter by adding `"/v2/apps/{APP_GUID}"`, `"GET \"/v2/jobs"`
+	- IP address - You may filter by adding `"ip: {IP_ADDRESS}"`
+	- vcap_request_id - You may filter by adding `"vcap-request-id:
+	  {REQUEST_GUID}"`
+
+You can combine many quoted attributes into one, by joining the quotes
+with `&&`, like so:
+
+```
+"for user: {GUID}" && "ip: {IP_ADDRESS}" && "GET \"/v2/jobs"
+```
+
+It probably won't end here, but it should provide a start for an
+individual to go further.
+
+## Getting further
+
+Once you found something interesting above, you could use it to refine
+your search to get deeper. For instance, one of the logs above, would have:
+
+- `@source.name` - Defining the VM we'd need to access. For example:
+  `api/0`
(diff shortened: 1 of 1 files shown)


docs/support/finding_activity.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Finding user GUID
+
+You may need to obtain this, to move the investigation forward. It's
+usually useful, to have a starting point in the form of organisation.
+Possessing that name, use this command, to obtain users from the
+organisation:
+
+```
+$ cf curl "/v2/users?q=organization_GUID:$(cf org ORG_NAME --guid)"
```
alext: I get the following for this:
```
$ cf curl "/v2/users?q=organization_GUID:$(cf org govuk-paas --guid)"
{
   "description": "The query parameter is invalid: organization_GUID",
   "error_code": "CF-BadQueryParameter",
   "code": 10005
}
```

alext: It needs the query param to be lowercase - `organization_guid`

docs/support/finding_activity.md
```diff
@@ -0,0 +1,48 @@
...
+We can make the investigation easier for ourselves by considering the
+following criteria:
+
+- Filter by user: `"for user: {GUID}"`
+- *Optionally:* Scope to job: `api`
+- `@message` should consist of:
+	- Detailed endpoint
+	- The response code, time
+	- IP address
+	- vcap_request_id
```
alext: It wasn't clear to me how to do this in Kibana, and took a bit of trial and error to figure out where to input this stuff. It would be good to make it very explicit how to construct this search.

docs/support/finding_activity.md
paroxp: Yes, sorry. I did a quick search and replace. Thanks.

docs/support/finding_activity.md
```diff
@@ -0,0 +1,48 @@
...
+
+We can make the investigation easier for ourselves by considering the
+following criteria:
+
+- Filter by user: `"for user: {GUID}"`
+- *Optionally:* Scope to job: `api`
+- `@message` should consist of:
+	- Detailed endpoint
+	- The response code, time
+	- IP address
```
dcarley: Does the IP address that's logged match the IP of the end user?

paroxp: Yes, I've checked that two days ago with my local IP whilst removing the whitelist.

docs/support/finding_activity.md
```diff
@@ -0,0 +1,103 @@
...
+
+It would essentially mean, that you can:
+
+```
+$ make prod bosh-cli
+...
+$ bosh ssh api/0
+$ sudo su -
+# cd /var/vcap/sys/logs
+# grep -R "{VCAP_REQUEST_ID}" . # Needs to be an actual GUID
```
dcarley: Are these also logged in Kibana? Can we provide queries so that we don't need to log into machines (which is slow and error prone)?

docs/support/finding_activity.md
```diff
@@ -0,0 +1,103 @@
...
+  `api/0`
+- `@message` consisting of `vcap-request-id`
+
+It would essentially mean, that you can:
+
+```
+$ make prod bosh-cli
+...
+$ bosh ssh api/0
+$ sudo su -
```
dcarley: If we need to keep these commands, can you remove this and use `sudo grep …` instead? We shouldn't advocate using `sudo su -` or variations that give an interactive root shell because it's easy to make mistakes and harder for other people to debug (because the commands aren't logged).

docs/support/finding_activity.md
dcarley: 👍

docs/support/finding_activity.md
paroxp: That was my initial thought, but I couldn't find these in Kibana :(

docs/support/finding_activity.md
```diff
@@ -0,0 +1,102 @@
...
+- `@message` consisting of `vcap-request-id`
+
+It would essentially mean, that you can:
+
+```
+$ make prod bosh-cli
+...
+$ bosh ssh api/0
+$ cd /var/vcap/sys/logs
+$ sudo grep -R "{VCAP_REQUEST_ID}" . # Needs to be an actual GUID
```
alext: Do you know which logfile it ends up finding these entries in?

I ask because I think we need a follow-up to figure out why this data isn't in Kibana.

paroxp: I may be eating my own words here...

Most of the results appear to come from the `./cloud_controller_ng/cloud_controller_ng.log` file. There are some logs in `./nginx_cc/nginx.access.log:api.cloud.service.gov.uk` - Although I don't think they're entirely relevant.

Now that I'm coming back to it, it looks like they're the same on both logs and kibana, which they weren't before. I suppose there is some sync period.

I'm happy to redefine the docs, to stay away from the CLI - What do you think?

alext: If we can get this information without having to SSH in to a server, that would be a big improvement. Please do update if that can be done.

paroxp: Upgraded. May want to re-review it.

PR #110, opened 2017-04-19
[#142318777] Update docs on CF upgrades

henrytk:
## What

[#142318777](https://www.pivotaltracker.com/n/projects/1275640/stories/142318777)

We haven't been following this documentation as our approach to picking
CF versions has relaxed over time with growing confidence. We change the
documentation to reflect our current approach to upgrades.

## How to review

Check for readability

## Who

Anyone but me


dcarley:
Thanks for doing this 👍

PR #112, opened 2017-04-21
[#144183633] Document the current process around CVES

timmow:
What
====

Document the process around CVES

How to Review
=============

Check the process makes sense

Who can review
==============

Anyone but me


dcarley:
👍  for writing this down

There's some duplication with other pages in the Team Manual that makes it less likely we'll read this new content. Would you be able to combine them?

- https://government-paas-team-manual.readthedocs.io/en/latest/team/cf_security_rss/
- https://government-paas-team-manual.readthedocs.io/en/latest/team/responding_to_security_issues/

timmow:
@dcarley most of what I wrote was already covered, i've moved where the docs live and made some very minor changes

docs/support/responding_to_security_issues.md
```diff
@@ -12,17 +12,19 @@ Periodically we will learn of a security issue affecting CloudFoundry or our sup
 
 1. The person on support should immediately create a Pivotal story with all the currently known information
 1. If a Deskpro ticket was raised, add the story link to the ticket and close the ticket
-1. The person on support should post a link to the story in #paas-incident and notify the main #the-government-paas channel that a security incident is in progress
+1. The person on support should notify the #the-government-paas channel that a security incident is in progress
```
dcarley: Could you describe the removal of `#paas-incident `? Is it because we don't treat these as incidents until we know something has happened? Should we still describe it as "a security incident is in progress"?

timmow: I removed it because I don't recall a CVE ever being posted in #paas-incident , so was trying to make the process docs match what actually happens. It also seems slightly overkill and that it could alarm people unnecessarily. On the other hand, it gives confidence that we are investigating the CVES

dcarley: We've never declared it a "incident incident" either. I think we should remove that wording, or move it further down as something we'd do once we've investigated and determined there might be a problem.

timmow: I've changed the wording

I thought about changing the wording to say we should declare a security incident, but if we have that in the docs we should then have a template of what we send tenants in a security incident, and what declaring a  security incident actually means - that seems like another PR

dcarley: 👍

PR #118, opened 2017-05-18
Make it easier to find users

paroxp:
## What

Once and if [paas-cf#925](https://github.com/alphagov/paas-cf/pull/925) will get merged, there will be a simpler way to
look for users in our CF environment. It's worth updating the docs for
future reference and promotion of the new script.

## How to review

- Have a read through
- Try to obtain user's GUID with the use of these docs


dcarley:
alphagov/paas-cf#925 was closed in favour of #119

PR #119, opened 2017-05-23
[SUPPORT] CF UAA user management guide

dcarley:
## What

Document how to do some things with `uaac`. This came out of discussion in alphagov/paas-cf#925

## How to review

1. Check that the words make sense
1. Try the commands

## Who can review

Ideally @paroxp, but anyone can.


timmow:
Running `bundle exec uaac token sso get cf -s ""` tells me to visit https://uaa.cloud.service.gov.uk/passcode - after visiting this I get a login form - after logging in and changing the redirect url I am sent to the getting started page that tenants see, so I don't get a passcode. Logging into cf works as expected.

docs/guides/CF_UAA_user_management.md
```diff
@@ -1,39 +1,88 @@
...
+
+UAA Command Line Interface
+…
+```
+
+### Log in
+
+To target an environment:
+
+    bundle exec uaac target https://uaa.<SYSTEM_DOMAIN>
```
timmow: This works for me if I target `https://login.<SYSTEM_DOMAIN>`

dcarley:
That's a pain. I think I might have been using a passcode URL from history rather than the one it prints. Good catch.

We could change the URL to `login.<SYSTEM_DOMAIN>` as you describe. It still points at UAA and everything else functions the same.

Or we could add an additional callback URL to Google to allow `uaa.<SYSTEM_DOMAIN>`.

What do you think?

timmow:
login is more consistent with the cf login process - that points you to get a token from `login.`

dcarley:
I've pushed it as an additional commit with an explanation because it's a bit weird.

docs/guides/CF_UAA_user_management.md
```diff
@@ -1,39 +1,88 @@
...
+
+There is a command to refresh your access token when it expires, but it is
+worth re-authenticating instead to ensure that you are still targeting the
+environment that you expect.
+
+## Finding a user account
+
+If you know the exact username of the account:
+
+    bundle exec uaac get <EMAIL>
```
bleach: I think this should be `bundle exec uaac user get <EMAIL>`

docs/guides/CF_UAA_user_management.md
```diff
@@ -1,39 +1,88 @@
...
-1. Get `uaac` command line via `gem install cf-uaac`
-2. Login with the admin _client_: `uaac token client get admin`. You will have to provide admin client secret as password.
-3. Get the user: `uaac user get <name@digital.cabinet-office.gov.uk>`. Note the user ID.
-4. Get properties of user in json form: `uaac curl /Users/<user_id> >u.json`
-5. Clean up the `u.json` file and keep only json properties of user.
-5. Edit user properties json - change `active` to `false`
-6. Update user:  `uaac curl /Users/<user_id> -H 'content-type: application/json' -X PUT -H 'If-Match: *' -d "$(<u.json)"`
-7. Verify changes applied by checking state: `uaac user get <name@digital.cabinet-office.gov.uk>`
+1. Get the user ID from [finding a user account](#finding-a-user-account).
+1. Get properties of user in json form: `uaac curl /Users/<user_id> >u.json`
```
bleach: I'm wondering if, for copy-paste convenience, we should add the `bundle exec` here and in the other uaac commands below?

docs/guides/CF_UAA_user_management.md
dcarley: Good catch, updated!

docs/guides/CF_UAA_user_management.md
dcarley: I wasn't sure whether to be that opinionated about Bundler. But yeah, I think we should, have updated in a separate commit.

PR #124, opened 2017-06-27
[#137364635] Update pull request merging guidance

henrytk:
## What

This is to reflect the fact we now sign merges into master on the
paas-bootstrap repository as well as paas-cf. Also, the GPG additions
to the git Concourse resource have long since been accepted upstream,
so we don't need to reference our old fork.

## How to review

* Check the links work.
* Review in conjunction with https://github.com/alphagov/paas-bootstrap/pull/73.

## Who

Anyone but me


alext:
Looks good to me.  Will merge when merging alphagov/paas-bootstrap#73

PR #127, opened 2017-07-11
[#145682111] Remove UAA SSO workaround guidance

henrytk:
## What

This is no longer necessary now that we are using uaa-release v41.

## How to review

Review in conjunction with https://github.com/alphagov/paas-cf/pull/971

Wait until https://github.com/alphagov/paas-cf/pull/971 is merged and deployed to CI and try to log into CI with your user account. Don't bother trying to test it in dev because it is a faff.

## Who

Only those with a CF admin user account associated with their GDS email address.


LeePorte:
Will merge during upgrade

bleach:
Noticed this wasn't merged. Since @LeePorte already reviewed it, it seems safe to do so.

dcarley:
May bad. Also 🐳 🐳 🐳

PR #129, opened 2017-07-18
[#145682111] Update CF upgrade docs for Datadog

henrytk:
## What

Multiple CF upgrades have now ran into problems with Datadog monitors
being broken.

## How to review

Proofread and check it makes sense.

## Who

Anyone but me

### DIFF
--- a/docs/guides/upgrading_CF,_bosh_and_stemcells.md
+++ b/docs/guides/upgrading_CF,_bosh_and_stemcells.md
@@ -22,7 +22,7 @@ git diff --exit-code -U5 --patience v228...v233 templates
 
 You should test the upgrade changeset:
 
-* From a fully deployed master, equivalent to the change that will happen in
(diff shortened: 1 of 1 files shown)


docs/guides/upgrading_CF,_bosh_and_stemcells.md
```diff
@@ -20,6 +20,8 @@ git diff --exit-code -U5 --patience v228...v233 templates
 
 ## Doing the upgrade
 
+Before deploying a new version make sure to enable Datadog and check for triggered monitors after deployment.
```
dcarley: I feel like this might be more likely to be read if it was a bullet point in the list below

henrytk: Done.

dcarley:
Thanks 👍

PR #131, opened 2017-07-25
Update standup time to be 9:45am

46bit:
Standup has been at 9:45am for at least 4 weeks (since I started.) The manual still says it is at 9:30am and this corrects the time.


alext:
👍  thanks for this.

PR #133, opened 2017-08-15
Update guidance on stemcell upgrades

timmow:
## What

Update stemcell Guidance

Stemcell upgrades are quite safe and it is good practice to apply the
stemcell upgrade while triaging the CVE. If the update applies cleanly
there is no reason we shouldn't upgrade the stemcell

## How to review

Check this makes sense, comment on if you think the split of stemcell upgrades and other security issues works

## Who can review

Anyone but me


dcarley:
👍

PR #136, opened 2017-08-25
[#149566861] ADR020: deletion of CI environment

combor:
## What

This ADR  explains why we decided to delete CI env, keep staging and
separate build CI

## How to review 

Read the content and decide if it's sane.

## Who

Not @combor

### DIFF
--- a/docs/architecture_decision_records/ADR020-deletion_of_ci_environment
+++ b/docs/architecture_decision_records/ADR020-deletion_of_ci_environment
@@ -1,0 +1,19 @@
+Context
+=======
+
+We have three environments in our deployment pipeline. Two non-production ones - CI and Staging and one Production. We think that it takes to much time for a change to reach production state in the current setup. We don't think having two environments before production is providing us much value, compared to the cost of running, maintaining, and waiting for deployments to be promoted.
+
+Decision
+========
+
+We will delete CI environment and migrate it's customizations, like tests , apps etc. to staging. We have decided to delete CI instead of staging as we want to separate build CI in it's own AWS account. Also, staging environment has valid certificates. 
+
+Status
+========
+
+Accepted.
+
+Consequences
+============
+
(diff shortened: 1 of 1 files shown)


docs/architecture_decision_records/ADR020-deletion_of_ci_environment
```diff
@@ -0,0 +1,19 @@
+Context
+=======
+
+We have three environments in our CI pipeline. Two non-production ones - CI and Staging and one Production. We think that it takes to much time for a change to reach production state in the current setup.
```
dcarley: Maybe "deployment pipeline" instead of "CI pipeline" because CI is also used to refer to an environment in this document?

docs/architecture_decision_records/ADR020-deletion_of_ci_environment
```diff
@@ -0,0 +1,19 @@
...
+
+Decision
+========
+
+We will delete CI environment. We don't think having two environments before production is providing us much value, compared to the cost of running, maintaining, and waiting for deployments to be promoted. We will delete CI environment and migrate it's customizations, like tests , apps etc. to staging. We have decided to delete CI as we want to separate build CI in it's own AWS account. Also, staging environment has valid certificates.
```
dcarley: Append to "We have decided to delete CI" + "instead of staging" to differentiate this from the decision to delete *an* environment?

dcarley: Should we move the 2nd sentence about cost into the "Context" section above where we talk about time-to-deployment so that all of the reasons for using thinking about this are in one place?

We can supplement this decision section with something to the effect of: we thought hard about this and couldn't see any reason not to do it.

combor:
Ok, fixed.

dcarley:
👍

PR #138, opened 2017-09-04
[#145189815] Moved BOSH connection scripts to be in the guide

46bit:
## What

We want to have and document a backup way to access bosh without running a task on concourse. In https://github.com/alphagov/paas-cf/pull/1023 we added several scripts to `paas-cf` for connecting to BOSH. In https://github.com/alphagov/paas-team-manual/pull/137 we added a guide to the team manual that told us how to use them.

The scripts have no automated tests, are liable to break in time, and may never be used. After more discussion in Pivotal [#145189815] we've decided to move the scripts to be code blocks in the Team Manual guide. This PR adds these code blocks, while https://github.com/alphagov/paas-cf/pull/1025 removes them from `paas-cf`.

Additionally, @leeporte identified issues with the scripts while approving:

* Configuring BOSH from scratch wasn't working. I've changed the script to work around this, but it does mean that BOSH credentials are output in the shell.
* Connecting to a temporary gateway VM with the state bucket's `id_rsa` was not working on their environment. I've worked around this by changing the script and instructions to use a new, temporary SSH keypair.

## How to review

* Attempt to follow the new instructions in the guide.

## Who can review

Not @46bit @henrytk @timmow


docs/guides/Connecting_to_Concourse_and_BOSH.md
```diff
@@ -50,16 +82,64 @@ Go into EC2 and create a new VM:
 * It may need to be on an `infra` subnet.
 * It must be assigned a public IP.
 * It must have both the `office-access-ssh` and `bosh-ssh-client` security groups. This gives SSH access from the GDS Office and SSH access to the BOSH VM.
-* It must use the `${DEPLOY_ENV}_key_pair` key pair for the script below to work.
-
-Use the `scripts/ssh_bosh_temporary_gateway.sh` script in `paas-cf` to SSH into the BOSH VM using this new VM as a gateway. You must first define the new VM's public IPv4 address as a `VM_IP` environment variable, for instance `VM_IP=53.33.78.122`.
+* It should be given a new key pair, which you can create (and download) in AWS.
```
dcarley: These instructions seem quite time consuming. I'm reluctant to re-test it, which makes me think I'd be unlikely to use it in an emergency. What do you think about omitting it and favouring "Emergency method: connecting directly to BOSH" instead?

docs/guides/Connecting_to_Concourse_and_BOSH.md
```diff
@@ -50,16 +82,64 @@ Go into EC2 and create a new VM:
...
+```
 
 You should terminate the new VM once you're done.
 
 ### Emergency method: connecting directly to BOSH
 
 If the Concourse VM is unavailable and creating new VMs is impossible or would take too long, you can make the BOSH VM accessible to our IP addresses. This should be a last-resort and temporary measure.
 
-Go into EC2 and add the `bosh/0` VM to the `office-access-ssh` security group. You should now be able to SSH to its public IP address using the `scripts/ssh_bosh_expose_vm.sh` script in `paas-cf`.
+Go into EC2 and add the `bosh/0` VM to the `office-access-ssh` security group. You should now be able to SSH to its public IP address using the below script.
```
dcarley: I don't understand this instruction. At first I thought it meant add (the IP of?) `bosh/0` to the inbound rules of `office-access-ssh`, but I don't think that would help, because BOSH isn't the client. Then I thought it might be add the `office-access-ssh` security group to the `bosh/0` VM, but it's not possible to modify the security group memberships of an existing VM. Do we need to modify `dcarley-bosh-ssh-client` (which BOSH is already a member of) instead?

docs/guides/Connecting_to_Concourse_and_BOSH.md
```diff
@@ -22,9 +22,41 @@ make dev bosh-cli
 
 If you can't use the task on Concourse, you can tunnel local `bosh` commands to the deployment's BOSH. 
 
-1. First install `bosh` with `gem install bosh_cli`.
-2. Second [establish an SSH connection to the BOSH VM](#ssh-into-the-bosh-vm).
-3. Third run the `scripts/bosh_cli_tunnel.sh` script in `paas-cf`. This forwards the BOSH port, logs you into BOSH and and drops you into a `bash` session to run `bosh` commands:
+1. First, install `bosh` with `gem install bosh_cli`.
```
dcarley: If you run `bundle exec bosh` from `alphagov/paas-cf` then it will use the version already specified in `Gemfile` which has the advantage that:

1. you don't need to install it separately - in my case I don't install gems to the root outside of Bundler
1. it uses a version that we know works

docs/guides/Connecting_to_Concourse_and_BOSH.md
```diff
@@ -22,9 +22,41 @@ make dev bosh-cli
...
+trap 'make dev stop-tunnel; rm -f /tmp/manifest.yml' EXIT
+
+echo "Making SSH tunnel to Bosh..."
+make dev tunnel TUNNEL=25555:10.0.0.6:25555
+
+echo
+BOSH_PASSWORD=$(aws s3 cp "s3://gds-paas-${DEPLOY_ENV}-state/bosh-secrets.yml" - | awk '/bosh_admin_password/ {print $2}')
+# bosh target has been asking for the username and password, but there is no way
+# to provide it from the command line except for bosh login.
+echo "Please log into BOSH with username 'admin' and password '$BOSH_PASSWORD':"
```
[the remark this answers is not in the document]
46bit: I'm using `BOSH 1.3262.26.0` and when running `bosh target` with a fresh install it asks for the username and password. If it isn't happening for you then it might be a bug in mine. What version are you using?

docs/guides/Connecting_to_Concourse_and_BOSH.md
46bit: I'm fine with that.

docs/guides/Connecting_to_Concourse_and_BOSH.md
46bit: You're right to find it confusing. It puts the target of the statement the wrong way round. It should be, "adding the `office-access-ssh` security group to the `bosh/0` vm." 

It is possible to modify security group memberships of an existing, running VM in the AWS Console. As part of this story I've checked that additions and removals take effect - seemingly immediately. Perhaps we should include the screenshot below in the Team Manual for where to find it?

screen shot 2017-09-06 at 09 55 47

dcarley: My mistake, you can do this now, that wasn't where I was expecting to find it. Either a screenshot or an `aws-cli` command would be good 👍

46bit: Good point, I'd forgotten the command for this might be simple.

docs/guides/Connecting_to_Concourse_and_BOSH.md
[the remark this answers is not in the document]
dcarley: Using the commands from https://github.com/alphagov/paas-cf/blob/d352f32626ab29f5bfea3f0345436eab1b4fc12d/concourse/scripts/bosh_login.sh#L10-L11 seems to work for both versions.

docs/guides/Connecting_to_Concourse_and_BOSH.md
```diff
@@ -22,9 +22,41 @@ make dev bosh-cli
 
 If you can't use the task on Concourse, you can tunnel local `bosh` commands to the deployment's BOSH. 
 
-1. First install `bosh` with `gem install bosh_cli`.
-2. Second [establish an SSH connection to the BOSH VM](#ssh-into-the-bosh-vm).
-3. Third run the `scripts/bosh_cli_tunnel.sh` script in `paas-cf`. This forwards the BOSH port, logs you into BOSH and and drops you into a `bash` session to run `bosh` commands:
+1. First, install `bosh` with `gem install bosh_cli`.
+2. Second, [establish an SSH connection to the BOSH VM](#ssh-into-the-bosh-vm).
```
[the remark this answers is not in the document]
46bit: We had misunderstood the `tunnel` target. It establishes a tunnel *via the Concourse VM*, so our instructions were misleading. I'm correcting things.

46bit:
I've committed a fixup which should cover all these issues, but I haven't yet checked the new instructions. I need to leave early today but I'll give that a go in the morning.

46bit:
@dcarley Ready for re-review.

dcarley:
Looks good 👍  I pushed a minor change to prefix the other two `bosh` commands with `bundle exec`. Could you rebase it in if you're happy? Then I'll merge.

PR #146, opened 2017-09-11
[#148989197] Move incident process into its own section

alext:
## What

Move the incident process into its own section so that it's easier to find in the event of an incident.

I've also fixed a fourth-wall link to no longer include Github Enterprise.

## How to review.

* Verify the changes make sense.
* Render locally (as described in the README) to ensure the navigation etc works correctly.

## Who can review

Not me.


dcarley:
👍

PR #151, opened 2017-10-02
[#149929104] Reference new uaa_lock_user.rb script

dcarley:
## What

Replaces the previous instructions that were quite difficult to use.

## How to review

Review with alphagov/paas-cf#1051

## Who can review

Anyone.

### DIFF
--- a/docs/guides/CF_UAA_user_management.md
+++ b/docs/guides/CF_UAA_user_management.md
@@ -67,14 +67,14 @@ If you only know part of the username:
 
 Sometimes it might be necessary to lock certain users. For example when we find out they have left GDS or don't have any more project to work on PaaS. CF has a facility to prevent users from logging in, while still preserving the user account with all their access rights and org membership. We do this as a first step in removing the user account. We then ask for confirmation from org managers (or managers of org managers in case we are removing an org manager) and only after confirmation finally remove the account completely (`cf delete-user`).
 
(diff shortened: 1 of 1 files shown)


docs/guides/CF_UAA_user_management.md
```diff
@@ -67,14 +67,14 @@ If you only know part of the username:
...
-1. Edit user properties json - change `active` to `false`
-1. Update user:  `bundle exec uaac curl /Users/<user_id> -H 'content-type: application/json' -X PUT -H 'If-Match: *' -d "$(<u.json)"`
-1. Verify changes applied by checking state: `bundle exec uaac user get <name@digital.cabinet-office.gov.uk>`
+From [alphagov/paas-cf](https://github.com/alphagov/paas-cf):
+```
+cd scripts
+bundle install
+TARGET=$(cf curl /v2/info | jq -r .token_endpoint) \
+  TOKEN=$(cf oauth-token) \
+  bundle exec ./uaa_lock_user.rb <USERNAME>
```
timmow: I dont know if this is my version of bundler, but I had to run

`bundle exec ruby ./uaa_lock_user.rb <USERNAME>`

For this to work

I also had to pass `--skip-ssl-validation` but its not worth adding this to the docs as it will mostly be used in production

dcarley: What was the error that you got when you ran it without?

timmow: Ahh no I got confused - the issue was with the missing `--skip-ssl-validation`, and I initally thought this was due to missing the ruby command

PR #152, opened 2017-10-12
Remove penalties

alext:
## What

After discussion at retro, it was agreed that the penalties weren't
constructive, and actually made some people feel less included. We
therefore decided that they should be removed.

I've therefore reworked the relevant sections to no longer mention any
penalties. I've also included them in the Orientation page as it makes
more sense for them to live there. I'll setup a redirect in the
ReadTheDocs admin console once this is merged.

This PR also includes some other refactorings to include the hand signals
page in the Orientation section, and remove an unused link.

## How to review

Read these changes. check they make sense.

## Who can review

Not me.


46bit:
I'm not sure Hand Signals is visible enough, but I don't like the idea of embedding big GIF(s). Up to you if you'd like to change that.

Otherwise, looks good to me.

alext:
> I'm not sure Hand Signals is visible enough, but I don't like the idea of embedding big GIF(s). Up to you if you'd like to change that.

I think if we want to change that, it should be a separate PR. This PR is focussed on removing the penalties. The other changes are merely moving things around.

46bit:
Good point. I suppose Hand Signals should have been a separate PR, as it previously had a dedicated page and now is amongst a variety of headings for Onboarding. I'll merge this and think about a later PR.

alext:
I've setup redirects in the readthedocs admin, although they don't work 100% because it's appending a / to the anchor.

PR #158, opened 2017-11-03
[#152150279] Describe API Latency procedures

paroxp:
## What

In case we need to respond to an incident caused by API Latency, we
would prefer to have a set of procedures in place to prevent us from
wasting time and look for the previous occurrences in Pivotal Tracker.

## How to review

- Check if the content makes sense

## Who can review

Neither @camelpunch nor myself.

### DIFF
--- a/docs/support/responding_to_alerts.md
+++ b/docs/support/responding_to_alerts.md
@@ -90,3 +90,34 @@ From `paas-cf` repository, run the following command:
 ```sh
 ./scripts/get-instance-details.sh "${RDS_INSTANCE_ID}"
 ```
+
+## API Latency
+
+Last time we received an alert for API latency, it was a UAA bug that caused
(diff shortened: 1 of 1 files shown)


docs/support/responding_to_alerts.md
```diff
@@ -90,3 +90,36 @@ From `paas-cf` repository, run the following command:
 ```sh
 ./scripts/get-instance-details.sh "${RDS_INSTANCE_ID}"
 ```
+
+## API Latency
+
+If a API Latency monitor is triggered due to an anomaly, we should aim to
```
samcrang: I don't really see the purpose of this sentence. I think it can just go?

camelpunch: Gone. I rewrote the following paragraph to suit - is it alright?

samcrang: Looks good 👍

PR #159, opened 2017-11-10
[#150711834] Update Bosh SSH instructions

henrytk:
## What

We now SSH to Bosh directly, so we can remove all instructions relating
to going via Concourse. We remove all of the backup options, because if
the Bosh VM isn't available then we have bigger problems than how to
use the Bosh CLI.

## How to review

Review in conjunction with https://github.com/alphagov/paas-bootstrap/pull/100 and https://github.com/alphagov/paas-cf/pull/1100
Probably best to merge this after those ones, as things could change as a result of the review.

* check it is accurate
* check it reads most goodly, with grammar correctly and stuff.
* run it locally using instructions in the README. Check it looks okay and functions properly.
* preferably run all instructions that have changed to see they still work.

## Who

Anyone but me or @camelpunch

### DIFF
--- a/docs/guides/Connecting_to_Concourse_and_BOSH.md
+++ b/docs/guides/Connecting_to_Concourse_and_BOSH.md
@@ -1,6 +1,6 @@
 ## SSH into the Concourse VM
 
-We can SSH directly into the Concourse VM using a Makefile target in `paas-cf`. This copies the Concourse VM's SSH key from a state bucket on S3, prints the VM's sudo password, and makes the SSH connection.
+We can SSH directly into the Concourse VM using a Makefile target in `paas-bootstrap`. This copies the Concourse VM's SSH key from a state bucket on S3, prints the VM's sudo password, and makes the SSH connection.
 
 ```
 make ssh_concourse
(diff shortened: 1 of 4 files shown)


henrytk:
I've just done a quick update, squashed commits, and force pushed. I've changed the instructions so you don't have to do `make dev bosh-cli`, you can do `DEPLOY_ENV=dev make bosh-cli`.

docs/guides/Connecting_to_Concourse_and_BOSH.md
```diff
@@ -10,111 +10,43 @@ make ssh_concourse
...
 ## Connecting bosh_cli to BOSH
 
-### Primary solution: use a Concourse task
-
-A Makefile task in `paas-cf` connects you to a one-off task in Concourse that's already logged into BOSH and has the deployment set using the CF manifest:
+A Makefile task in `paas-bootstrap` pulls a Docker image and runs a container loaded with the secrets required to interact with Bosh using the CLI:
 
 ```
-make dev bosh-cli
+DEPLOY_ENV=dev make bosh-cli
```
alext: I think this is a little confusing because it's unlikely that the `DEPLOY_ENV` is `dev`, and this might suggest that the `DEPLOY_ENV` should be set to the `AWS_ACCOUNT` value.

It's probably clearer to use `foo` as an example `DEPLOY_ENV`

henrytk: I agree. Done. Squashed. Pushed.

docs/guides/IPSec_debugging.md
```diff
@@ -1,7 +1,7 @@
 ## IPSec debugging
 
 
-There may be a need, to debug the IPSec on our machines. In order to do that, you may want to log into `bosh-cli` and then `bosh ssh {router|cell}`.
+There may be a need, to debug the IPSec on our machines. In order to do that, you may want to log into the routers or cells. See [Connecting bosh_cli to Bosh](https://government-paas-team-manual.readthedocs.io/en/latest/guides/Connecting_to_Concourse_and_BOSH/#connecting-bosh_cli-to-bosh) for instructions.
```
alext: It would be better to use a relative link here. This means that the links will work when running locally etc.

henrytk: Sensible. Done.

PR #163, opened 2017-11-16
Add Travis config to run linkchecker against the built site

alext:
## What

Run linkchecker against the built site so that we catch any broken links etc.

This fixes a couple of broken links that were found as a result.

I've also cleaned up an unused image that was missed in #159

## How to review

* Code review
* Verify that the Travis build passes.

## Who can review

Not me.


requirements.txt
```diff
@@ -18,3 +18,7 @@ six==1.10.0
 tornado==4.3
 webcolors==1.5
 wsgiref==0.1.2
+
+linkchecker==9.3
```
samcrang: The `linkchecker` project currently seems to be in a weird state—[supposedly the project is dead](https://github.com/wummel/linkchecker/issues/649#issuecomment-340248279) and has been [picked up by a new group](https://github.com/linkcheck/linkchecker/) but they still [haven't released anything on PyPI yet](https://github.com/linkcheck/linkchecker/issues/4).

Does this using this seemingly dead project seem like a problem?

alext: We're already using this for the tech docs, so I chose to use it here for consistency. I agree that it's in a wierd state, but I think we should address that across everything separately if that becomes an issue.

PR #167, opened 2017-11-28
Add guide for releasing BOSH blobs

bandesz:
## What

We had to add a new release blob to a service broker but we had authentication problems with the BOSH cli. We documented the process so others have clear instructions how to do this.

## How to review

Read through the guide and check for spelling mistakes.

## Who can review

Not @bandesz


alext:
:+1: thanks for adding this.

PR #173, opened 2018-01-03
[#153367232] Add notes about gorouter stdout and stderr logs

bandesz:
## What

Add notes about how to find gorouter stdout and stderr logs.

## How to review

Read it, check the formatting.

## Who can review

Not @bandesz


dcarley:
Thanks 👍

PR #182, opened 2018-02-02
Describe current process for story approval

camelpunch:
# What

The team manual seems to reflect an older process where a single Product Manager would approve stories. This change updates the manual to reflect the current process where almost any team member can approve.

Also, adjust work-in-progress limits to match those of Rubbernecker.

# Why

It's confusing for new / prospective team members to read about a process that we don't use.

# Who can't review

@camelpunch


camelpunch:
Rubbernecker UI currently has a 6 underneath "Doing". Does this mean it's limited to 6, or has no limit but happens to have 6 in flight?

PR #187, opened 2018-02-23
[#120268519] Monitoring gorouter latency

LeePorte:
## What

This PR adds methods for investigating and dealing with high latency alerts
to the team manual. I expect that this section should be uexpanded upon
after team discussion.

## How to review

Checkout this branch and run `mkdocs serve` then visit http://127.0.0.1:8000/support/responding_to_alerts/ to make sure the content renders properly and makes sense.

## Who can merge?

No @LeePorte

### DIFF
--- a/docs/support/responding_to_alerts.md
+++ b/docs/support/responding_to_alerts.md
@@ -270,3 +270,13 @@ The following Trusted Advisor checks cause expected warnings:
 * S3 Bucket Permission's for `gds-paas-*-ci-releases-blobs` are required to be publicly accesible so it is safe to mark these as excluded.
 * ELB Listener Security for `*-ssh-proxy`: This ELB is in TCP mode and so this warning is expected
 * Security Group - Specified Ports Unrestricted for `*-sshproxy`: We intentionally allow port 2222 for `cf ssh` access so it is safe to mark this as excluded.
+
+## Gorouter high latency alerts
+
+If you see an alert for gorouter latency being high;
+
+* Check with the team if we have any known load testing occuring either by the team or by a tenant
+* Visit [kibana](https://logsearch.cloud.service.gov.uk/) and filter by `@source.component: gorouter` then add in `gorouter.host` to the table, this will show you which hosts are being most used currently
+* Contact the team who own the service to see if they are aware of anything happening
+
+As this is the first monitor of this type please investigate the gorouters to discover the issue they are encountering. We have previously seen high resource usage (CPU and Memory) these should be checked in the first case.


docs/support/responding_to_alerts.md
```diff
@@ -270,3 +270,21 @@ The following Trusted Advisor checks cause expected warnings:
...
 * Security Group - Specified Ports Unrestricted for `*-sshproxy`: We intentionally allow port 2222 for `cf ssh` access so it is safe to mark this as excluded.
+
+## Gorouter high latency alerts
+
+If you see an alert for gorouter latency being high;
+
+* Check with the team if we have any know load testing occuring
+* Visit [kibana](https://logsearch.cloud.service.gov.uk/app/kibana#/discover?_g=(refreshInterval:(display:Off,pause:!f,value:0%29,time:(from:now-15m,mode:quick,to:now%29%29&_a=(columns:!(gorouter.host%29,filters:!(('$state':(store:appState%29,meta:(alias:!n,disabled:!f,index:'logstash-*',key:'@source.component',negate:!f,value:gorouter%29,query:(match:('@source.component':(query:gorouter,type:phrase%29%29%29%29%29,index:'logstash-*',interval:auto,query:(query_string:(analyze_wildcard:!t,query:'*'%29%29,sort:!('@timestamp',desc),vis:(aggs:!((params:(field:'@source.name',orderBy:'2',size:20%29,schema:segment,type:terms),(id:'2',schema:metric,type:count%29%29,type:histogram%29)%29&indexPattern=logstash-*&type=histogram) this will show you which hosts are being most used currently
```
dcarley: This link doesn't render correctly in Markdown. When I copy/paste it into a browser then Kibana has a red banner at the top that says "Unable to parse URL".

We could link to a saved search but I'd be worried that we might lose/recreate, as we did recently when we deleted the `.kibana` index when looking at a problem.

Could you detail the search term and fields that you need to look at instead?

docs/support/responding_to_alerts.md
```diff
@@ -270,3 +270,21 @@ The following Trusted Advisor checks cause expected warnings:
 * S3 Bucket Permission's for `gds-paas-*-ci-releases-blobs` are required to be publicly accesible so it is safe to mark these as excluded.
 * ELB Listener Security for `*-ssh-proxy`: This ELB is in TCP mode and so this warning is expected
 * Security Group - Specified Ports Unrestricted for `*-sshproxy`: We intentionally allow port 2222 for `cf ssh` access so it is safe to mark this as excluded.
+
+## Gorouter high latency alerts
+
+If you see an alert for gorouter latency being high;
+
+* Check with the team if we have any know load testing occuring
```
dcarley: There's a typo, `know` -> `known`.

It might also be worth explicitly saying that it could be a tenant load testing their app on PaaS. Otherwise it could be interpreted as checking whether someone in our team is running a load test.

docs/support/responding_to_alerts.md
```diff
@@ -270,3 +270,21 @@ The following Trusted Advisor checks cause expected warnings:
...
+
+Options friom here are to restart the loadbalancers
+
+```sh
+cd ${PAAS_BOOTSTRAP_DIR}
+DEPLOY_ENV=prod make prod bosh-cli
+bosh restart router # You may want/need to target an instance, e.g. router/23888ebf-2dd3-4afe-b370-51705403d423
+```
+
+Or to scale up the load balancers if the CPU and Memory metrics indicate that this is a bottleneck
```
dcarley: Should we say "up" (vertically) or "out" (horizontally)?

docs/support/responding_to_alerts.md
LeePorte: Interesting was that using the github preview or via `mkdocs serve`?

Using the latter I see the following.

responding to alerts gov uk paas team manual

dcarley: Ahh, it was GitHub's preview. I hadn't tested with mkdocs yet.

I still think it's worth changing because the URL is quite complex and it's possible that it might break if we upgrade Kibana. It would be good to have a simple way to reproduce the same results.

docs/support/responding_to_alerts.md
LeePorte: I was leaning towards up as we would only have one change in this case rather than having to have multiple changes to account for more ip address usage.

docs/support/responding_to_alerts.md
```diff
@@ -270,3 +270,21 @@ The following Trusted Advisor checks cause expected warnings:
...
+
+## Gorouter high latency alerts
+
+If you see an alert for gorouter latency being high;
+
+* Check with the team if we have any know load testing occuring
+* Visit [kibana](https://logsearch.cloud.service.gov.uk/app/kibana#/discover?_g=(refreshInterval:(display:Off,pause:!f,value:0%29,time:(from:now-15m,mode:quick,to:now%29%29&_a=(columns:!(gorouter.host%29,filters:!(('$state':(store:appState%29,meta:(alias:!n,disabled:!f,index:'logstash-*',key:'@source.component',negate:!f,value:gorouter%29,query:(match:('@source.component':(query:gorouter,type:phrase%29%29%29%29%29,index:'logstash-*',interval:auto,query:(query_string:(analyze_wildcard:!t,query:'*'%29%29,sort:!('@timestamp',desc),vis:(aggs:!((params:(field:'@source.name',orderBy:'2',size:20%29,schema:segment,type:terms),(id:'2',schema:metric,type:count%29%29,type:histogram%29)%29&indexPattern=logstash-*&type=histogram) this will show you which hosts are being most used currently
+* Contact the team who own the service to see if they are aware of anything happening
+
+Options friom here are to restart the loadbalancers
```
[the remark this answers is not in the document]
LeePorte: If tenant apps are taking a long time to respond then it will keep threads open on the router resulting in less capacity to process new incoming requests.

docs/support/responding_to_alerts.md
dcarley: It feels weird that we instinctively scale up instead of out, but I don't feel that strongly about it.

docs/support/responding_to_alerts.md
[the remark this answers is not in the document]
dcarley: I'd prefer that we remove the advice because it supposes a particular situation and requires context to understand whether it will make it better or worse.

I think we should restrict runbook actions to things that we know will be safe to do, because that's probably the assumption that the reader will be operating under.

docs/support/responding_to_alerts.md
```diff
@@ -270,3 +270,21 @@ The following Trusted Advisor checks cause expected warnings:
...
+
+## Gorouter high latency alerts
+
+If you see an alert for gorouter latency being high;
+
+* Check with the team if we have any known load testing occuring either by the team or by a tenant
+* Visit [kibana](https://logsearch.cloud.service.gov.uk/) and filter by `@source.component: gorouter` then add in `gorouter.host` to the table, this will show you which hosts are being most used currently
+* Contact the team who own the service to see if they are aware of anything happening
+
+Options friom here are to restart the loadbalancers
```
alext: Why do we think that restarting the routers will help?

Secondly, I think the term 'loadballancers' is confusing here. I initially read this as referring to the ELBs, which we don't have control over. Maybe this should be 'routers' as that's what the VMs are called.

docs/support/responding_to_alerts.md
alext: I agree, I can see that before long we'll want routers in all 3 AZs - it'll mean we need less headroom to handle deploys etc.

docs/support/responding_to_alerts.md
dcarley: The thread has been collapsed because the line has been modified but there's a similar conversation going on here: https://github.com/alphagov/paas-team-manual/pull/187#discussion_r170569528

PR #189, opened 2018-03-01
Emphasise that our Working Practices aren't set in stone

46bit:


dcarley:
Just noticed that this has changed from WIP. Merging!

PR #190, opened 2018-03-01
It's OK to pair in any way that you like

46bit:


alext:
I agree, this is good. @46bit any reason for this to still be marked WIP?

dcarley:
Just noticed that this has changed from WIP. Merging!

PR #191, opened 2018-03-02
[Support] Update UAAC instructions for new Gemfile layout

alext:
We no longer have a separate Gemfile for the scripts directory[1]. This
updates the insteuctions here accordingly.

[1]https://github.com/alphagov/paas-cf/commit/42e9fc88e6bc812a36dfeb3890e5373320170d58


keymon:
LGTM

PR #200, opened 2018-05-01
[#156658598] Allow google to index the site.

alext:
## What

We want to use a Google site-scoped search box on this site. For that to work we need Google to index it.

This also adds a google site verification file so that it can be added to the Search Console

## How to review

* Run this locally
* Verify that there is no `<meta name="robots" content="noindex">` tag in the `<head>`
* Verify that `http://localhost:4567/google52ece46ed12b8c21.html` returns the file.

## Who can review

Not me.

### DIFF
--- a/config.rb
+++ b/config.rb
@@ -2,6 +2,8 @@ require 'govuk_tech_docs'
 require 'html-proofer'
 
 activate :directory_indexes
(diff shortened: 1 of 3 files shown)


keymon:
:+1:

PR #213, opened 2018-05-16
Expand IA and CO SIRO acronyms

dcarley:
## What

So that people can understand what they are because:

- IA has other meanings such as Information Architecture
- CO and SIRO aren't commonly used

## How to review

Check that it makes sense.

## Who can review

Anyone.


bleach:
:clap:

PR #218, opened 2018-05-24
Zendesk has replaced DeskPro

dcarley:
## What

Replace all occurrences in the Team Manual and:

- we don't have any equivalent content for setting up Zendesk, probably
  because most of the setup was already done by the team that manages
  it, so delete the page.
- we now use a Google Group for the PagerDuty integration.

## How to review

Check the accuracy of the content and whether it renders correctly.

## Who can review

Anyone.


bandesz:
Superseeded by https://github.com/alphagov/paas-team-manual/pull/226

PR #219, opened 2018-05-24
Replace names of Product Managers with role

dcarley:
## What

Tom no longer works on the team. Refer to the people by their roles, so
that we don't need to update this for subsequent team changes.

## Why

Check the rationale and content.

## How to review

Anyone.


urmishah:
Makes sense. Merging this

PR #220, opened 2018-05-24
Update story process

dcarley:
## What

To reflect the fact that:

- review and approval can be done by the same person
- we no longer have a Business Analyst
- we have Feature Leads
- we haven't been using work in progress limits
- names on stories indicate availability of work

I've collapsed some of the existing content together as a result.

## How to review

Check that it makes sense and reflects reality.

## Who can review

Anyone.

### DIFF
--- a/source/team/working_practices.html.md.erb
+++ b/source/team/working_practices.html.md.erb
@@ -199,6 +199,10 @@ git push origin master
 
 # Stories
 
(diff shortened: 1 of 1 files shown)


urmishah:
This reflects what we have agreed. Merging

dcarley:
👋 👍

urmishah:
@dcarley 👋

PR #230, opened 2018-06-20
 Update description of our review process.

alext:
## What

We've agreed a change in our process so that when pairing on a story, a
pair can review and merge their own work. This updates the manual to
reflect this.

I've also restored some content from the README around generating diagrams that got removed inadvertently.

## How to review

Verify this makes sense and matches what we agreed.

## Who can review

Not me.


source/team/working_practices.html.md.erb
```diff
@@ -233,9 +231,25 @@ it must indicate follow up work that is required. Listing options uncovered duri
 the spike is good. Indicate which option is preferable. Where possible follow up stories
 should be noted. Doing this allows the Product Manager prioritise the follow up work.
 
-## Review and approval
+## Review
```
chrisfarms: I think maybe this is complicating things more than required... the only real rule I see is "at least two people must be involved in the process" .... maybe we could simplify further, something like:

# Development process

* Stories go through 3 stages:
    * **Doing** actively working on implementing the story. Doing is complete once the owner(s) are happy with their implementation.
    * **Reviewing** gathering feedback and testing. Reviewing is complete once the owner(s) are satisfied the "it's done when" from the story is met and any revelevent code is merged.
    * **Approving** [optional] sometimes there are actions that must be taken after a change has made it out to production or it might be requested in the story that a particular person double checks the changes are perfoming as expected, stories can be left in "Approving" for these actions to be completed.
* All stories **must be worked on by at least two people**
* Two people may take their story through all 3 stages or may opt to leave a story in one of the further stages in-order to get feedback for more developers.

alext: > Two people may take their story through all 3 stages or may opt to leave a story in one of the further stages in-order to get feedback for more developers.

When we discussed this, we decided against this - ie the approver must always be someone different from the people who worked on the story.

alext:
Following discussion with @chrisfarms I've updated the approval section to say that this should be done by the Feature or Tech Lead. This ensures that the feature lead can keep track of
the progress of stories in a given epic.

PR #231, opened 2018-07-04
Commands to get ext IPs and RDS DBs for pentest

keymon:
When requesting pen-tests approvals to AWS we also shall send
the external IPs of our cluster (NAT instances, I guess) and the
RDS instances affected.

We assume that only the RDS instances for the given tenant org
will be affected.

### DIFF
--- a/source/guides/Tenant_Application_Penetration_Testing.html.md.erb
+++ b/source/guides/Tenant_Application_Penetration_Testing.html.md.erb
@@ -64,3 +64,20 @@ The `cdn-route` service uses a CloudFront distribution. You can discover these d
 aws cloudfront list-distributions \
 | jq '.DistributionList.Items[] | select(.Comment == "cdn route service") | .Id' --raw-output
 ```
+
+## Find external IPs (NAT)
+
+```
+aws ec2 describe-nat-gateways | jq -r '.NatGateways[].NatGatewayAddresses[].PublicIp
+```
+
+## Find RDS instances from a given org
+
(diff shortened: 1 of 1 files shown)


source/guides/Tenant_Application_Penetration_Testing.html.md.erb
```diff
@@ -64,3 +64,20 @@ The `cdn-route` service uses a CloudFront distribution. You can discover these d
 aws cloudfront list-distributions \
 | jq '.DistributionList.Items[] | select(.Comment == "cdn route service") | .Id' --raw-output
 ```
+
+## Find external IPs (NAT)
+
+```
+aws ec2 describe-addresses | jq  -r '.Addresses[] | select ( .InstanceId == null ) | .PublicIp'
```
samcrang: If you're looking specifically for NAT gateway public IPs you can:

```
aws ec2 describe-nat-gateways | jq -r '.NatGateways[].NatGatewayAddresses[].PublicIp'
```

source/guides/Tenant_Application_Penetration_Testing.html.md.erb
```diff
@@ -64,3 +64,20 @@ The `cdn-route` service uses a CloudFront distribution. You can discover these d
...
+```
+
+## Find RDS instances from a given org
+
+Assuming you are logged in the production CF and or Ireland
```
samcrang: Do you mean "and/or London"?

keymon:
@samcrang updated

PR #235, opened 2018-07-09
ADR028 - decision of moving platform logs to logit

venusbb:
## What

Add ADR028 - the Decision to continue moving platform logs to Logit.io

## How to review

Proofread. Make sure it makes sense and reflects our decision and the implications. You can run it locally to check the link is correct.

## Who

Anybody but me


alext:
I think the content that's here looks good, however, it's missing the "Consequences" section.

alext:
Looks good to me. I've squashed the commits and will merge as soon as CircleCI is done.... Oh it's done before I finished typing this message.... Merging.

PR #237, opened 2018-07-13
[#158943533] Add ADR for Aiven project setup

richardTowers:


alext:
Gah!, I've just realised that there's already an ADR028 in this PR: https://github.com/alphagov/paas-team-manual/pull/235. We should update this to use 029

alext:
I've rebased this to renumber it to ADR029.

richardTowers:
Thanks!

venusbb:
I question the need for creating individual members. We will only create security benefit, if some of the team members can log in Aiven as 'operator' but don't have access to our cred store; or some team members who can access to Aiven as 'admin' but don't have access to the high cred store. 

When a team member leave, other than deleting the individual's Aiven's member account, we still need to rotate the Aiven's team credentials.  This is because if they can access the cred store or high cred store, they will have the details to access the Aiven account.

venusbb:
Also, please confirm to PaaS PR template.

richardTowers:
Updated to follow the PR template. I've also added [a template](https://github.com/alphagov/paas-team-manual/blob/master/.github/PULL_REQUEST_TEMPLATE.md) in a separate PR so that future PRs in this repo will follow this format by default.

alext:
> I question the need for creating individual members. We will only create security benefit, if some of the team members can log in Aiven as 'operator' but don't have access to our cred store; or some team members who can access to Aiven as 'admin' but don't have access to the high cred store.

We do have some team members who who don't have access to the cred store, but do need access to Aiven. Additionally, there are very few users who have access to the high credstore, and this set doesn't necessarily align with those who need admin access to Aiven.

The main reason for having per-user accounts is to encourage good practice, and to make it easier to have an audit trail of who's done what. It also allows for people to setup MFA on their logins.

Ideally, the per-project users wouldn't exist, and we would be able to define API keys at a project level instead of them being attached to a user. This isn't currently possible due to how Aiven structure things. We could raise a feature request about this.

PR #238, opened 2018-07-16
[#158938536] Simplify the docs on updating CF / fix formatting

richardTowers:
What
----

Describe who can review the changes. Or more importantly, list the people
that can't review, because they worked on it.

It looks like Middleman's markdown parser needs more space before the
third level indent for some reason, so the "main driver" bits were showing
at the second level instead of the third (see [the current output](https://team-manual.cloud.service.gov.uk/guides/upgrading_CF,_bosh_and_stemcells/#upgrading-cf-bosh-and-stemcells)).

I found the instructions for picking a version to be tricky to
understand - they seem to say "the latest version (only sometimes not
because they're broken (only even if they're broken you can just fix
them))" which seems overcomplicated. I've updated them to be more
opinionated about picking the latest stable release and fixing bugs in
`cf-deployment`.

I've also made a couple of URLs links, as the markdown parser doesn't
make them clickable otherwise.

Finally I've also snuck a PULL_REQUEST_TEMPLATE in.

How to review
-------------

Check that the new instructions make sense.

Who can review
--------------

Anyone who knows about upgrading cloud foundry.

### DIFF
--- a/source/guides/upgrading_CF,_bosh_and_stemcells.html.md.erb
+++ b/source/guides/upgrading_CF,_bosh_and_stemcells.html.md.erb
@@ -4,13 +4,12 @@
 
 * Separate the upgrade of Cloud Foundry and stemcells from the upgrade of Bosh. Upgrades can cause problems and our experience is that it is difficult to be certain about the cause of those problems if multiple things have changed.
 * Establish the correct version to upgrade to:
-  * This will usually be the latest stable release, or the one before it, subject to review.
-  * The main driver for not always picking the latest release:
-    * This is to avoid the bad releases which CF produce from time to time.
-    * An alternative strategy may be to pick a release at least one week old, thereby giving Pivotal time to flag and pull bad releases before we plough ahead with them.
-    * This is less of a requirement since `cf-deployment` as specific releases versions can be easily changed to fix bugs.
-  * Kick-off is an appropriate place for this review.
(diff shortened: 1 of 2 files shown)


source/guides/upgrading_CF,_bosh_and_stemcells.html.md.erb
```diff
@@ -4,13 +4,12 @@
...
-    * This is to avoid the bad releases which CF produce from time to time.
-    * An alternative strategy may be to pick a release at least one week old, thereby giving Pivotal time to flag and pull bad releases before we plough ahead with them.
-    * This is less of a requirement since `cf-deployment` as specific releases versions can be easily changed to fix bugs.
-  * Kick-off is an appropriate place for this review.
-  * Check [cf-deployment releases documentation](https://github.com/cloudfoundry/cf-deployment/releases) before upgrading to a particular version.  There have been releases that became inappropriate for production usage.
+  * Check [cf-deployment releases documentation](https://github.com/cloudfoundry/cf-deployment/releases).
+  * Prefer the latest stable release if the release notes look like they won't introduce bugs.
+  * Discuss the planned upgrade version in kick-off.
+
+* If you encounter small bugs during the upgrade consider fixing them on our fork of `cf-deployment`.
```
keymon: I would rephrase this. We will avoid change `cf-deployment`. Instead, `cf-deployment` refers to some releases, but we can override them with an opsfile.

I would say something in the lines of:

> If you encounter issues in the releases refer by `cf-deployment` you might consider fork and patch them or pick override the release version with a opsfile.

richardTowers:
Ok, will squash the first two commits before merging. FYI I've also added a pull request template to this PR.

PR #246, opened 2018-09-07
[support] Update links to logging stack

LeePorte:
What
----

We are no longer running our own ELK stack, as a result the links were
out of date. They have been corrected to the new links.

How to review
-------------

Code review and run locally with `bundle exec middleman server` and visit http://localhost:4567/support/responding_to_alerts/

Who can review
--------------

Not @LeePorte

### DIFF
--- a/source/support/responding_to_alerts.html.md
+++ b/source/support/responding_to_alerts.html.md
@@ -9,9 +9,11 @@ it may help to look at UAA logs.
 
 The access logs is on the UAA VMs in `/var/vcap/sys/log/uaa/localhost_access.log`.
 
-All other logs are sent to [logsearch](https://logsearch.cloud.service.gov.uk):
+All other logs are sent to [logit](https://logit.io/a/1c6b2316-16e2-4ca5-a3df-ff18631b0e74).
+You must [Log into logit via Google](https://reliability-engineering.cloudapps.digital/manuals/logit-io-joiners.html)[external link] before you can access kibana.
 
 * Scope to component: UAA
+* filter by `@source.deployment: "prod"` or `@source.deployment: "prod-lon"` (depending on which deployment you are looking for)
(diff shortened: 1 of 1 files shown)


source/support/responding_to_alerts.html.md
```diff
@@ -9,7 +9,7 @@ it may help to look at UAA logs.
 
 The access logs is on the UAA VMs in `/var/vcap/sys/log/uaa/localhost_access.log`.
 
-All other logs are sent to [logsearch](https://logsearch.cloud.service.gov.uk):
+All other logs are sent to [logit](https://logit.io/a/1c6b2316-16e2-4ca5-a3df-ff18631b0e74):
```
alext: This link will only work if you're already logged into Logit. Otherwise it'll start you down the wrong login flow, which won't work. I think we need to restructure this page a little, and link to the RE instructions for logging into Logit.

source/support/responding_to_alerts.html.md
```diff
@@ -215,7 +215,7 @@ The following Trusted Advisor checks cause expected warnings:
 If you see an alert for gorouter latency being high;
 
 * Check with the team if we have any known load testing occuring either by the team or by a tenant
-* Visit [kibana](https://logsearch.cloud.service.gov.uk/) and filter by `@source.component: gorouter` then add in `gorouter.host` to the table, this will show you which hosts are being most used currently
+* Visit [kibana](https://kibana.logit.io/s/665ca355-efc3-46a2-96cf-21d31a5305bb/app/kibana#/) and filter by `@source.component: gorouter` then add in `gorouter.host` to the table, this will show you which hosts are being most used currently
```
alext: Again, this link will only work if you're already logged into Logit.

Also while we're here, we should probably add a filter for `@source.deployment`, otherwise you'll get logs from all deployments sharing that Logit stack.

PR #259, opened 2018-10-17
Document creating new users / orgs

richardTowers:
What
----

Document how we create new orgs so people on support know what to do

How to review
-------------

Review the instructions are actually what we should do

Who can review
--------------

Anyone


richardTowers:
I'm going to rethink the script since paazmin should be the way to add users now - the script should just be for adding new orgs.

PR #261, opened 2018-10-22
Credhub upstream divergence 160506139

jpluscplusm:
What
----

Added a section to the CF upgrade page calling out our decision not to include upstream's `credhub` instance group, and the implications that'll have for future upgrades.

How to review
-------------

Check the words make sense and are in a suitable place.

Who can review
--------------

@alext is a likely candidate.


jpluscplusm:
[ #gitpocalypse2018: don't merge until https://github.com/alphagov/paas-team-manual/pull/261/commits/cbabe6e664e1ddecf7d0ebd4bcf5eab45f903aa8 is a stable and refreshable page, and GH's UI stops bouncing around between different commits or 404'ing ]

PR #273, opened 2018-12-14
[#162612712] Remove Datadog references from ADRs

alext:
What
----

We're removing Datadog, so these no longer apply. Instead of updating
them to refer to the equivalent metrics in the new system, I've reworded
them to generally describe the thing being measured so that they're
agnostic of the metrics system being used, and therefore shouldn't need
updating again if this changes.

I've also added a link to ADR032 from ADR021 as it supersedes one detail of it.

How to review
-------------

Check the new wording makes sense.

Who can review
--------------

Not me.

### DIFF
--- a/source/architecture_decision_records/ADR021-cell-capacity-assignment-2.html.md
+++ b/source/architecture_decision_records/ADR021-cell-capacity-assignment-2.html.md
@@ -94,7 +94,7 @@ Proposed
 We will alert on the following metrics:
 
 - exponentially weighted moving average of used memory (not including cache) averaged over cells > 50%
-- smoothed idle CPU averaged across cells < 50%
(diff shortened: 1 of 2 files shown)


source/architecture_decision_records/ADR021-cell-capacity-assignment-2.html.md
```diff
@@ -94,7 +94,7 @@ Proposed
 We will alert on the following metrics:
 
 - exponentially weighted moving average of used memory (not including cache) averaged over cells > 50%
-- smoothed idle CPU averaged across cells < 50%
+- smoothed idle CPU averaged across cells < 50% (Note: This metric superceeded by [ADR32](/architecture_decision_records/ADR023-idle-cpu-alerting-change/))
```
LeePorte: Typo s/32/23/

alext: Fixed...

PR #283, opened 2019-03-19
[#163521334]Warn about resource starvation in dev during upgrades

AP-Hunt:
What
----
During CF upgrades, resource starvation of a dev environment can lead to
unexpected test failures. It's easy enough to resolve, but not obvious that it
happens.

How to review
-------------

1. Check the content
2. Check it renders correctly

Who can review
--------------
Anyone but me

### DIFF
--- a/source/guides/upgrading_CF,_bosh_and_stemcells.html.md
+++ b/source/guides/upgrading_CF,_bosh_and_stemcells.html.md
@@ -18,7 +18,9 @@
   git diff v1.0.0...v1.14.0 cf-deployment.yml
   ```
 
-  We also use a number of upstream ops files, so you will want to `diff` them too. See [the manifest generation script](https://github.com/alphagov/paas-cf/blob/master/manifests/cf-manifest/scripts/generate-manifest.sh) for which ones get used.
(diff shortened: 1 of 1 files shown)


source/guides/upgrading_CF,_bosh_and_stemcells.html.md
```diff
@@ -18,7 +18,9 @@
   git diff v1.0.0...v1.14.0 cf-deployment.yml
   ```
 
-  We also use a number of upstream ops files, so you will want to `diff` them too. See [the manifest generation script](https://github.com/alphagov/paas-cf/blob/master/manifests/cf-manifest/scripts/generate-manifest.sh) for which ones get used.
+* Check your environment isn't resource starved, because this can cause unexpected test failures. Typically, resource starvation in dev happens when there are a number of organisations left over from smoke and acceptance tests. These are prefixed with `SMOKE-` and `CATS-` respectively. Provided tests aren't running at the time, they can be safely deleted with `cf delete-org`.
+
+We also use a number of upstream ops files, so you will want to `diff` them too. See [the manifest generation script](https://github.com/alphagov/paas-cf/blob/master/manifests/cf-manifest/scripts/generate-manifest.sh) for which ones get used.
```
richardTowers: I feel like this paragraph belongs with the previous bullet (about git diff-ing the cf-deployment) - it's got spliced away by your addition.

AP-Hunt: Good call. Will rearrange.

AP-Hunt:
I've moved the new bullet point to the top of the list, because it felt like putting it anywhere else was disrupting an otherwise easy to follow set of steps

PR #284, opened 2019-03-28
Update incident process to restore original meaning

tnwhitwell:
What
----

The Tech writer review appears to have changed the meaning of this line - from "this may be the person on support, it now reads as "this must be the person on support"

I have restored the original meaning of the line

How to review
-------------

Read the new wording, and confirm it is equivalent to the original intention of the line.

Who can review
--------------

Not @tnwhitwell. May require tech writer 2i (@jonathanglassman)


richardTowers:
One word change, so I don't think it's worth putting it through 2i.

PR #285, opened 2019-03-28
[#161684547] Add nosy neighbour info to manual

LeePorte:
What
----

We have enabled the noisy neighbour nozzle, this commit adds a page about what it is and how to use the nozzle.

How to review
-------------

Run it locally and make sure it makes sense

Who can review
--------------

anyone

### DIFF
--- a/source/support/finding_noisy_apps.html.md
+++ b/source/support/finding_noisy_apps.html.md
@@ -1,0 +1,12 @@
+# Finding noisy apps
+
+We have had situations on the PaaS where a noisy neighbour issue has caused problems in the log stack and for other tenants. Identifying these can be hard.
+
+We have enabled the noisy-neighbour-nozzle to aid in locating these.
+
+Use;
+
+1. Install the cf cli plugin `cf install-plugin -r CF-Community "log-noise"`
+1. Run cf log-noise this will give you the top 10 noisiest logging apps on the platform
+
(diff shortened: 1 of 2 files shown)


source/support/finding_noisy_apps.html.md
```diff
@@ -0,0 +1,12 @@
+# Finding noisy apps
+
+We have had situations on the PaaS where a noisy neighbour issue has caused problems in the log stack and for other tenants. Identifying these can be hard.
+
+We have enabled the noisy-neighbour-nozzle to aid in locating these.
+
+Use;
+
+1. Install the cf cli plugin `cf install-plugin -r CF-Community "log-noise"`
+1. Run cf log-noise this will give you the top 10 noisiest logging apps on the platform
```
bandesz: There is a step missing that you should target the admin org and the monitoring space.

```
cf target -o admin -s monitoring
```

source/support/finding_noisy_apps.html.md
```diff
@@ -0,0 +1,12 @@
...
+
+If you are looking for a particular app you can use `log-noise nn-nozzle nn-accumulator <app name>`
```
bandesz: This makes no sense, the help suggests you only need to use a parameter if your app name is not `nn-accumulator`.

```
NAME:
   log-noise - Show top log producers from noisy-neighbor-nozzle accumulator.

USAGE:
   log-noise <nozzle accumulator app name>
```

PR #289, opened 2019-04-12
ADR 31 - separate jobs in cloudfoundry pipeline

tlwr:
What
----

Add an ADR for changes to the pipeline

How to review
-------------

Agree with the ADR

Who can review
--------------

Not @tlwr

### DIFF
--- a/source/architecture_decision_records/ADR031-services-core-pipeline-separate-jobs.html.md
+++ b/source/architecture_decision_records/ADR031-services-core-pipeline-separate-jobs.html.md
@@ -1,0 +1,41 @@
+# ADR031: Separate PaaS services from the Platform core pipeline
+
+
+# Context
+
+We have a single pipeline `create-cloudfoundry` which creates a Cloud Foundry
+deployment, and also deploys additional services to the platform.
+
+These services include:
+
+- PaaS Accounts
+- PaaS Admin
+- PaaS Billing
+- PaaS Metrics
+
+Which are core components to our platform, but not to Cloud Foundry.
+
+Currently these services are unnecessarily coupled in a couple of places:
+
+- The `post-deploy` job
+- The `custom-acceptance-tests` job
+
+Unnecessarily coupling has resulted in flakey app tests blocking CVE
+remediation from reaching production.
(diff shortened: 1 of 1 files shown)


source/architecture_decision_records/ADR031-services-core-pipeline-separate-jobs.html.md
```diff
@@ -0,0 +1,38 @@
...
+deployment to the core platform.
+
+
+# Status
+Accepted
+
+
+# Consequences
+
+The pipeline will no longer be fully controlled by the `pipeline-lock` pool.
```
richardTowers:
```suggestion
The services we've separated out will deploy as soon as self-update-pipelines changes the tag_filter of their resource. They will also deploy for each new version of paas-cf. This means they will deploy twice (once when the pipeline-lock job finishes and again when post-deploy finishes).
```

source/architecture_decision_records/ADR031-services-core-pipeline-separate-jobs.html.md
```diff
@@ -0,0 +1,38 @@
...
+These services include:
+
+- PaaS Accounts
+- PaaS Admin
+- PaaS Billing
+- PaaS Metrics
+
+Which are core components to our platform, but not to Cloud Foundry.
+
+Currently these services are unnecessarily coupled in a couple of places:
```
AP-Hunt: Could you document the negative side-effects of the coupling? I think the context would benefit from further explanation of the problem this ADR is addressing.

tlwr: Would something like

> Unnecessarily coupling has resulted in flakey app tests blocking CVE remediation from reaching production.

be what you had in mind?

AP-Hunt: Yep, lgtm.

tlwr: Added

PR #296, opened 2019-05-22
Diagram of Certificate Authority rotation

46bit:
What
----

I drew this diagram to analyse whether we really had to do quite so many steps to rotate our CAs. I found a shorter path (the thick arrows) but don't plan to implement it right now because of the time-consuming testing that would be needed.

People have said the diagram is helpful. It's only a first draft but this commit puts it in to the page to help visual thinkers (like me!)

How to review
-------------

Do you like it?

Who can review
--------------

Not @46bit


tlwr:
Miki starts a Zine, Soon ™️ :)

46bit:
😂 I have about 17 mental health zines at home 😄 but maybe not suitable to publish 😦

PR #303, opened 2019-07-09
Record the decision to continuously deploy CF apps

richardTowers:
What
----

Add an ADR to record the decision to change the way the CF apps
(paas-admin, paas-billing and paas-cf) are deployed by the pipeline.

How to review
-------------

* Check the ADR is correct and understandable

Who can review
--------------

Not Rich


tlwr:
This looks like a coherent representation of consensus within the team

PR #309, opened 2019-09-02
Clarify user management

tlwr:
What
----

After the SSO work we have state split across multiple microservices, in
a way which is difficult to reconcile.

We don't delete users when we do leavers tickets, instead just lock them
and remove all their user permissions.

This should be in the team manual.

How to review
-------------

Read it

### DIFF
--- a/source/support/CF_UAA_user_management.html.md
+++ b/source/support/CF_UAA_user_management.html.md
@@ -86,7 +86,9 @@ The endpoint also works for UAA GUIDs:
 
 ## Locking a user account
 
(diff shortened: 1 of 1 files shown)


richardTowers:
Agree that this should be in the team manual, but I'm not sure that locking users when they leave is the right option. Hanging on to email addresses we don't need puts us in a not-great position re:GDPR.

tlwr:
I agree RE GDPR but I would argue we can justify it because

a) we only store the user's email address
b) a guid is not sufficient for attributing actions within the platform, we probably want a piece of PII for this

If this isn't the case then we should:

a) add a `DELETE /users/:guid` endpoint to paas-accounts
b) add [Delete user] functionality to paas-admin which removes the user from
- CF
- UAA
- PaaS accounts

And then update this documentationm

PR #311, opened 2019-09-03
Correct author’s name in link to blog post

quis:
The author is _Julia_ Evans, at least according to https://jvns.ca/about/ and https://twitter.com/b0rk


richardTowers:
Oof, how embarrassing. Sorry Julia!

Thanks for catching this Chris!

PR #313, opened 2019-09-04
Add ADR447 - new RDS broker plans

richardTowers:
What
----

Adds an architectural decision record to create some new database plans.

How to review
-------------

* Review the ADR
* Once approved, send it back to @richardtowers to raise the follow up
  stories and update the ADR with links.

Who can review
--------------

Not @richardtowers

### DIFF
--- a/source/architecture_decision_records/ADR447-add-new-rds-broker-plans.html.md
+++ b/source/architecture_decision_records/ADR447-add-new-rds-broker-plans.html.md
@@ -1,0 +1,171 @@
+Recommended database plans
+==========================
+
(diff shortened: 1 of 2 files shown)


tlwr:
Big 👍 from me, someone else from the team should probably also have a look

I think the analysis done by @AP-Hunt was 🔝 ; as a nominee 😈

AP-Hunt:
This all looks and sounds sensible to me. I've forced-pushed the commit to include a link to the ADR on the index page, with the other ADRs.

Approving.

PR #328, opened 2019-12-05
[#169870405] Add ADR for aiven metrics for users via prometheus

tlwr:
[Story](https://www.pivotaltracker.com/story/show/169870405)

What
----

Adds an ADR which records some of the decisions from the spike

How to review
-------------

Read the ADR

Post comments and questions for things I have not sufficiently explained or things that do not make sense

Who can review
--------------

@tlwr and @paroxp worked on the story, and so other people should review


tlwr:
Added a new commit clarifying why Caddy and Gorouter are used in this way.

AP-Hunt:
Do we need to/should we include in here something about how we're going to manage the multi-tenancy of the data?

tlwr:
@AP-Hunt I would assume that is covered by:

> In the initial implementation we will use Caddy and Gorouter to load balance Prometheus. This should be seen as a temporary implementation which will be revisited when we look at how to expose the Prometheus API to users.

unless I am misunderstanding your comment

source/architecture_decision_records/ADR450-aiven-metrics-for-users.html.md
```diff
@@ -0,0 +1,50 @@
...
+- ship logs to an Elasticsearch/Rsyslog
+- send metrics to Datadog
+- send metrics to Aiven Postgres/InfluxDB
+- expose metrics in Prometheus exposition format
+
+We currently run Prometheus for monitoring the platform, using the [Prometheus BOSH release](https://github.com/bosh-prometheus/prometheus-boshrelease) and have confidence and experience using it.
+
+We will need to think about Prometheus failover. If we load balance Prometheus without sticky sessions, then the metrics reported by Prometheus will be erratic as different instances report different metrics.
+In the initial implementation we will use Caddy and Gorouter to load balance Prometheus. This should be seen as a temporary implementation which will be revisited when we look at how to expose the Prometheus API to users.
+Caddy does automatic failover, such that during normal operation all traffic gets proxied to a single instance, but when the normal instance is unreachable, the local Prometheus is used.
```
46bit: What is "the local Prometheus" in this context? Should this say "an alternate Prometheus" instead?

46bit:
I suspect Andy was thinking about how these metrics are for multiple tenants and we should restrict access to them to the appropriate tenant. I share those concerns, but there's nothing here about how data is exposed from the Prometheus to tenants. I think I'm OK to leave that for now.

tlwr:
I've pushed up a new commit with a diagram, which renders as:

image

source/architecture_decision_records/ADR450-aiven-metrics-for-users.html.md
tlwr: Good point, I have reworded this in the latest commit, specifically change this to "colocated prometheus" and the diagram should demonstrate what _colocated_ means

tlwr:
I have also reworded the "this architecture is temporary" sentence to maybe address Andy's point better.

PR #331, opened 2020-01-13
Document the Operator and Admin roles

richardTowers:
What
----

Document the new operator role, how it differs from the Admin role, and
when to use which role.

How to review
-------------

* Check the docs make sense
* No need for tech writer 2i, as this is internal documentation

Who can review
--------------

Not @richardtowers


tlwr:
:shipit:

PR #332, opened 2020-01-28
[#170894236] Add ADR 451

LeePorte:
What
----

This commit adds the new ADR taken on how to move forwards with removing the final shared credentials in use as part of PCI compliance and better securing the platform.

How to review
-------------

Spin it up locally and check that everything renders correctly.
Read the ADR and check it makes sense.

Who can review
--------------

Anyone

### DIFF
--- a/source/architecture_decision_records/ADR451-bosh-access-without-socks.html.md
+++ b/source/architecture_decision_records/ADR451-bosh-access-without-socks.html.md
@@ -1,0 +1,43 @@
+# BOSH access without SOCKS5 or SSH
+
+## Context
+
+To comply with [PCI Requirements](https://www.pcisecuritystandards.org/pci_security/maintaining_payment_security) we will remove the use of shared credentials anywhere in the platform.
+
+We will use Single Sign On (SSO) to avoid credential sharing.
+
+To enable SSO we will expose some of the APIs on the BOSH instance without using a SOCKS5 proxy. This is due to the SOCKS5 proxy not playing nicely with the VPN, which in turn creates a support risk
+
+Discussion has taken place with both IA and with Cyber on the proposed methods after reviewing the [RFC created as part of #169915408](https://docs.google.com/document/d/1XZsrNp88tOSyC_bjy1mg3Yyv2TkpKgYSjoYResGAbps/edit#heading=h.xscqoqxlc072)
+
+## Decision
+
+We will remove the reliance on SOCKS5 or SSH tunnels for the UAA API endpoint so that we can remove the use of shared credentials.
+
+With these in place we are unable to complete an SSO journey from one of our existing IDPs as the BOSH instance is not broswer accessible.
+
+The current method will be replaced with Single Sign On in order to obtain a UAA token. This token is then used to interact with the rest of the APIs on the BOSH instance.
+
+## Implementation
+
+In the initial implementation we will place UAA on the bosh director in the internet using the IP Whitelist and using Google for SSO
+
+Access to both BOSH and Credhub will remain via SSH Tunnels during this phase of implementation.
+
+![architecture](../images/adr451-bosh-access-without-socks.svg)
+
+By using this method we are retaining the benefit of using IP whitelisting whilst removing the shared credential that is in use.
+
+## Status
+
+Accepted
+
+## Consequences
+
+We will no longer rely on IP Whitelisting or SSH tunnels and have individual credentials for accessing UAA on the BOSH Director.
+
+Full alerting can be undertaken by Cyber for this component.
+
+We will not be using SSH tunnels to obtain UAA tokens.
+
+
--- a/source/architecture_decision_records/ADR452-bosh-access-with-mtls.html.md
+++ b/source/architecture_decision_records/ADR452-bosh-access-with-mtls.html.md
@@ -1,0 +1,45 @@
+# BOSH access with mTLS
+
+## Context
+
+In [ADR451 BOSH access without SOCKS](../ADR451-bosh-access-without-socks/) we removed the requirement for using a SOCKS5 proxy or SSH tunnel to access UAA.
+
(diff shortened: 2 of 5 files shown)


howardstaple:
As discussed, IA are content with this approach. IA review of implementation will be through PaaS SWG

source/architecture_decision_records/ADR451-bosh-access-changes.html.md
```diff
@@ -0,0 +1,52 @@
+# BOSH access changes
+
+## Context
+
+We have the requirement to no longer use shared credentials anywhere within the platform in order to satisfy PCI requirements. The only place where we are left doing this is on the BOSH director.
+
+In order to migrate away from the useage of shared credential and to using Single Sign On (SSO) we need to expose some of the API's on the BOSH director to the internet in order for the SSO journey to complete.
```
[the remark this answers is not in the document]
pauldougan: A suggested rewording of the context section to use an more active voice and add a reference link to the PCI requirements

```
Currently we share credentials to access the BOSH director.

To comply with [PCI Requirements](https://www.pcisecuritystandards.org/pci_security/maintaining_payment_security) we will remove the use of shared credentials anywhere in the platform.

We will use Single Sign On (SSO) to avoid credential sharing.

To enable SSO we will expose some of the BOSH director API’s  to the internet.
``

Comment (tlwr) [inline source/architecture_decision_records/ADR451-bosh-access-changes.html.md]:
```suggestion
In order to migrate away from the useage of shared credential and to using Single Sign On (SSO) we need to expose some of the API's on the BOSH instance without a SOCKS5 proxy in order for the SSO journey to complete.
```

s/bosh directory/bosh instance ; director is a job on the bosh instance

Also I would like to avoid using the term _to the internet_ because the intent is ambgiuous

source/architecture_decision_records/ADR451-bosh-access-changes.html.md
```diff
@@ -0,0 +1,52 @@
...
+
+Discussion has taken place with both IA and with Cyber on the proposed methods after reviewing the [RFC created as part of #169915408](https://docs.google.com/document/d/1XZsrNp88tOSyC_bjy1mg3Yyv2TkpKgYSjoYResGAbps/edit#heading=h.xscqoqxlc072)
+
+## Decision
+
+We will remove the reliance on IP Whitelisting and the use of shared credentials.
+
+These will be replaced with Mutual TLS and with Single Sign On
```
tlwr: I think we could split this into two ADRs, as we have not bottomed out the discussion of the security ramifications of using mTLS to replace managed IP safelists

source/architecture_decision_records/ADR451-bosh-access-changes.html.md
```diff
@@ -0,0 +1,52 @@
...
+
+We will take a multiple implementation step approach in order to build confidence in how this works and to retain the security of the platform.
+
+## Initial implementation
```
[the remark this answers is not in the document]
tlwr: As above, I think this should be two separate ADRs, the mTLS ADR referencing the former "expose UAA ADR"

tlwr:
Nice graphviz diagrams :D

Summarising my comments above, I think this should be two ADRs

- Expose UAA without a SOCKS proxy, which details why the SSO journey needs this
- Use mTLS instead of a managed IP safelist (this should also be threat modelled properly - which we haven't done e.g. denial of service of BOSH instance) (@olliejc)

source/architecture_decision_records/ADR451-bosh-access-changes.html.md
```diff
@@ -0,0 +1,54 @@
...
+
+Access to both BOSH and Credhub will remain via SSH Tunnels during this phase of implementation.
+
+This architecture is temporary and will be revised in the second implementation step.
+
+![architecture](../images/adr451-bosh-access-changes-step1.svg)
+
+## Follow up implementation
+
+Mutual TLS will be implemented on the BOSH Director in order to gain access to the APIs contained.
```
richardTowers: Another argument for pulling this into a separate ADR I think, but it would be nice to go into a bit more detail about how this will be achieved.

tlwr:
Thanks for splitting these into two ADRs

I think ADR 451 can be accepted and implemented straight away

I think ADR 452 needs to be re-reviewed by IA and Cyber because I do not think they have covered the denial of service vector, and potential vulnerabilities introduced by removing the IP safelist (AWS security group acting as a packet filter). ADR 452 should take into account or 

Exposing the APIs on the BOSH director without an IP filter introduces two threat vectors:

- Denial of service
  1. An attacker running `vegeta attack` against the component (which is colocated on the BOSH director) which attempts to negotiate a mTLS attack, can use shared resources (CPU, Memory, Network IO & sockets) to prevent the BOSH director from doing its job
  1. An attacker using [slow loris](https://www.cloudflare.com/learning/ddos/ddos-attack-tools/slowloris/) attacks, initiating connections slowly, and attempting to hold them open. This could prevent the BOSH director from being able to initiate outbound connections to the instances that it manages

- Credential leak via memory over/underflow, misc networking vulnerabilities. The component doing mTLS is (under assumptions in this ADR) colocated on the same instance as BOSH, which has privileged access and lots of credentials/passwords/private keys. A vulnerability such as [Heartbleed](https://en.wikipedia.org/wiki/Heartbleed) (historical OpenSSL vulnerability) would be present a fairly high risk.

---

I think the approach outlined in ADR 452 has been sanctioned because of an incorrect assumption about how mTLS would be implemented, i.e. having the component which does mTLS be a separate VM or component from the BOSH director.

CCing @howardstaple @olliejc @akinnane

howardstaple:
Not sure I completely agree on why it was sanctioned. IA were content with the approach but the ADR contains no details on how mutual auth TLS would be implemented. The expectation is that a robust but pragmatic approach to implementation would be taken, where a risk needed to be accepted then we would get sign off to do so or get told to find another way.

LeePorte:
With ADR 452 we are committing to a direction without having fully undertaken the work, hence it is  guiding structure at this point. I suggest that it is committed in that vein with a story acceptance criteria to add an updated ADR with the agreed and implemented method of achieving the direction in ADR 452

PR #341, opened 2020-03-11
[#171180051] Document how to rotate broker passwords

46bit:
What
----

CVE-2020-5400 leaked the basic auth credentials of some our brokers. We played Pivotal Tracker story #171180051 to rotate those credentials. It added a new pipeline job. Unlike our other credential rotation processes this one involved downtime because our brokers lack support for changing basic auth credentials without downtime.

This commit documents the process so that we retain knowledge. I'm most keen to have documented what this new job doesn't do: it only rotates the basic auth credentials, not anything else. That is important context to retain.

How to review
-------------

Read through the documentation changes and see if you have all your questions answered.

Who can review
--------------

Not @46bit.

PR #98, opened 2017-02-27
[#130020741] Document how to rotate credentials

mtekel:
## What
Document how to rotate credentials using the tasks introduced in https://github.com/alphagov/paas-cf/pull/802

## Review
Proofread. Check it makes sense.

## Who
not @mtekel

PR #342, opened 2020-03-11
Refresh how to rotate the platform credentials

46bit:
What
----

The documentation on how to rotate platform credentials had not been substantially updated since it was written in #98 in 2017. In that time we've moved to Credhub and done various other things.

This commit refreshes the documentation, tries to improve clarity, and adds sensible warnings and context.

The strongest motivation for this change (other than doing #341 to this page at the same time) was that _"First, you delete credentials using one of the CI helper jobs."_ is _wrong_ now. We never delete credentials from Credhub, we regenerate them.

How to review
-------------

Compare the old and new documentation. Do you think the new form is clearer? Give feedback.

Who can review
--------------

Not @46bit.

### DIFF
--- a/source/team/rotating_credentials.html.md
+++ b/source/team/rotating_credentials.html.md
@@ -9,33 +9,77 @@ We store credentials for third-party services in [`pass`](https://www.passwordst
 
 ## Platform
 
-From time to time, it might be necessary to rotate our platform credentials. We have created helper tasks in the deployer and bootstrap pipelines, credentials tab, to clean passwords that we know are safe to re-set.
+From time to time, it might be necessary to rotate our platform credentials. We
+have created helper tasks in the deployer and bootstrap pipelines to regenerate
+passwords that we know are safe to be rotated. This can be found under the
+`credentials` tab of each Concourse pipeline.
 
-### Rotating passwords
+### Rotating platform credentials
 
-Rotating passwords consists of two phases. First, you delete credentials using
-one of the CI helper jobs.
+We have Concourse jobs for rotating credentials that can be rotated without
+downtime. These might not include everything. If you are responding to a
+security incident then it is worth checking whether the credentials you
+care about are included.
 
-There are two for CF (in the deployer pipeline), one for BOSH and one for
-Concourse (in the bootstrap pipeline). These tasks exclude passwords that are
-difficult or not safe to rotate and clean all other passwords. We do also
-delete generated ssh keys in these tasks.
+Credential rotation is done in two phases:
 
-There are two CF cred rotation jobs (`rotate-cf-admin-password` and
-`rotate-cloudfoundry-credentials`, under
-`/teams/main/pipelines/create-cloudfoundry?groups=credentials`) so that you
-have the choice of rotating the admin password or not. Rotating the admin
-password automatically leads to the rest of the credentials being rotated, but
-not vice versa.
+1. rotation in the secret store (Credhub)
+2. deployment of the secrets
 
-Second phase consists of running the main pipeline (on deployer for CF, on bootstrap for BOSH and Concourse) to generate new credentials and apply them. Ensure that there is nothing else triggering this 2nd phase run, as that would mean you would be rotating passwords _and_ trying to apply some changes at the same time, which might not work and you could end up with broken deployment. Pipeline run should complete successfully and all tests should pass. There should be no interruptions to deployed apps.
+These jobs are not run very often. It is wise to re-examine them for drift. Perform a test run in a development environment before executing them in a production environment.
+
+If something goes wrong then in many cases
+you can obtain the old passwords from the credential history in Credhub using
+the `--versions` flag of `credhub get`:
+
+```
+$ credhub get -n /prod-lon/prod-lon/cf_admin_password --versions=10
+
+- id: 764582ca-b3f0-4d08-8da4-6a48b69907fd
+  name: /prod-lon/prod-lon/cf_admin_password
+  type: password
+  value: [redacted newest password]
+  version_created_at: "2020-03-13T11:22:50Z"
+- id: 949f97e8-5ba3-4e64-8bf2-49dc3b94ec7a
+  name: /prod-lon/prod-lon/cf_admin_password
+  type: password
+  value: [redacted previous password]
+  version_created_at: "2020-03-11T10:10:13Z"
+```
+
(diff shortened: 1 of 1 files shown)


source/team/rotating_credentials.html.md
```diff
@@ -9,26 +9,41 @@ We store credentials for third-party services in [`pass`](https://www.passwordst
...
+passwords that we know are safe to re-set. This can be found under the
+`credentials` tab of each Concourse pipeline.
 
-### Rotating passwords
+### Rotating platform passwords
 
-Rotating passwords consists of two phases. First, you delete credentials using
-one of the CI helper jobs.
+We have two jobs for rotating platform passwords that can be rotated without
+downtime. These could not include all the passwords so it is worth digging in
```
AP-Hunt: Can we include notable exceptions?

source/team/rotating_credentials.html.md
```diff
@@ -9,26 +9,41 @@ We store credentials for third-party services in [`pass`](https://www.passwordst
 
 ## Platform
 
-From time to time, it might be necessary to rotate our platform credentials. We have created helper tasks in the deployer and bootstrap pipelines, credentials tab, to clean passwords that we know are safe to re-set.
+From time to time, it might be necessary to rotate our platform credentials. We have created helper tasks in the deployer and bootstrap pipelines to clean
```
AP-Hunt: What does "clean" mean in this context?

source/team/rotating_credentials.html.md
```diff
@@ -9,26 +9,41 @@ We store credentials for third-party services in [`pass`](https://www.passwordst
...
+We have two jobs for rotating platform passwords that can be rotated without
+downtime. These could not include all the passwords so it is worth digging in
+if we ever have a security incident.
 
-There are two for CF (in the deployer pipeline), one for BOSH and one for
-Concourse (in the bootstrap pipeline). These tasks exclude passwords that are
-difficult or not safe to rotate and clean all other passwords. We do also
-delete generated ssh keys in these tasks.
+There are two phases to rotating passwords. First you run a job to rotate the
+credentials where they are stored (e.g., in Credhub.) Second you run the full
```
AP-Hunt: s/.,/./

source/team/rotating_credentials.html.md
```diff
@@ -9,26 +9,41 @@ We store credentials for third-party services in [`pass`](https://www.passwordst
...
+deploy pipeline to apply the new credentials.
```
AP-Hunt: We should use the name of the pipeline (`create-cloudfounry`) here, rather than referring to is as "the full deploy pipeline". It's worth being unambiguous.

source/team/rotating_credentials.html.md
```diff
@@ -9,26 +9,41 @@ We store credentials for third-party services in [`pass`](https://www.passwordst
...
+* `rotate-prometheus-credentials`
+* `rotate-database-encryption-keys`
+
+Before running each of these jobs you should make sure no deploy is in
+progress. After running one of these jobs you must trigger a full run of
+its pipeline in order to deploy the rotation.
+
+Since these jobs are not run very often it is wise to re-examine them a bit
+for safety before triggering them. If something goes wrong then in many cases
+you can obtain the old passwords from the credential history in Credhub.
```
AP-Hunt: If something has gone wrong, I'm likely going to be looking at this page. It'd be good to have a quick sample of how to get history from Credhub.

source/team/rotating_credentials.html.md
```diff
@@ -9,26 +9,41 @@ We store credentials for third-party services in [`pass`](https://www.passwordst
...
+credentials where they are stored (e.g., in Credhub.) Second you run the full
+deploy pipeline to apply the new credentials.
 
-There are two CF cred rotation jobs (`rotate-cf-admin-password` and
-`rotate-cloudfoundry-credentials`, under
-`/teams/main/pipelines/create-cloudfoundry?groups=credentials`) so that you
-have the choice of rotating the admin password or not. Rotating the admin
-password automatically leads to the rest of the credentials being rotated, but
-not vice versa.
+In `create-bosh-concourse` there are two rotation jobs under the `credentials`
```
AP-Hunt: Something about these two paragraphs makes me uneasy. I think I'd like to expand a bit more on what the individual jobs do, and whether I can or can't run any of them one after another without running a deploy between.

source/team/rotating_credentials.html.md
```diff
@@ -9,26 +9,41 @@ We store credentials for third-party services in [`pass`](https://www.passwordst
...
+
+In `create-cloudfoundry` there are three rotation jobs under the `credentials`
+tab:
+
+* `rotate-cloudfoundry-credentials`
+* `rotate-cf-admin-password`
+* `rotate-prometheus-credentials`
+* `rotate-database-encryption-keys`
+
+Before running each of these jobs you should make sure no deploy is in
```
AP-Hunt: We should also pause the pipeline so that deployments don't start while we're not looking.

source/team/rotating_credentials.html.md
46bit: I've rephrased this a little and moved it around. I don't know what might not be included, but I want people to be aware that they should double-check whether the things they care about are included in these rotation jobs. It's a shame that people have to make that call when using them, but I don't see a way to ensure that these jobs are always kept up to date.

source/team/rotating_credentials.html.md
46bit: This applied to both pipelines. I've restructured a bit to try and address this.

source/team/rotating_credentials.html.md
46bit: Added a nice example.

source/team/rotating_credentials.html.md
46bit: This was an odd word choice on my part. It should have been "rotate" or "regenerate."

source/team/rotating_credentials.html.md
46bit: Perhaps we should pair on that next week. I get your point, but I'm nervous of writing guarantees in that might not be updated when the underlying code gets changed.

46bit:
Brilliant review Andy. I've marked as resolved the comments that I reckon I've uncontroversially addressed. Let's go over the others next week.

46bit:
Paired on fixes with Andy :)

PR #351, opened 2020-06-03
[#173150927] Guidance on applying tenant ElastiCache service updates

tlwr:
[Story](https://www.pivotaltracker.com/story/show/173150927)

What
----

Did some cleanup on the index page to separate out support guidance and how-tos

Made ADR links easier to read

Added guidance on how to conduct elasticache (redis) service updates

How to review
-------------

Code review

Review locally

Check the following screenshots

Screenshot of new index page
<caption>Screenshot of new index page</caption>

Screenshot of new content
<caption>Screenshot of new guidance</caption>

Who can review
--------------

A developer who will apply service updates


source/support/tenant_elasticache_service_updates.html.md
```diff
@@ -0,0 +1,92 @@
...
+
+For highly available service instances, downtime will be limited, but for
+non-highly available instances, there may be some downtime (up to 30 minutes).
+
+## When should we upgrade
+
+Depending on the update, it might be more or less important to upgrade.
+AWS emails PaaS Support when updates are available.
+
+Also we may wish to do schedule an update when a many instances have an
```
AP-Hunt, line 18: Suggest

> We may also wish to schedule an update when many instances have an update available. You can determine this percentage with the Prometheus query:

source/support/tenant_elasticache_service_updates.html.md
```diff
@@ -0,0 +1,92 @@
...
+count(paas_aws_elasticache_cluster_update_required)
+/
+count(paas_aws_elasticache_replication_group_nodes_count)
+```
+
+## Process outline
+
+1. Create a Pivotal story to track the lifecycle of the upgrade
+
+1. Pick maintenance windows, one earlier, one later
```
AP-Hunt, line 31: Are these deliberately before and after office hours? If so, we should probably call that out.

source/support/tenant_elasticache_service_updates.html.md
```diff
@@ -0,0 +1,92 @@
+# Applying tenant ElastiCache service updates
```
AP-Hunt, line 1: Not sure where this should live within the document, hence line 1.

I think we should make it clearer that applying the updates is still a manual process. It would be easy to think that the maintenance windows set in the tool will cause the updates to take place.

source/support/tenant_elasticache_service_updates.html.md
tlwr: Resolved in 52f92c1

source/support/tenant_elasticache_service_updates.html.md
tlwr: Added a caveat

source/support/tenant_elasticache_service_updates.html.md
tlwr: Added a caveat

PR #372, opened 2020-09-17
Add accessibility statement

cmcnallygds:
What
----

Add accessibility statement and link to it in footer

How to review
-------------

1. Preview locally and make sure content renders properly
2. Check that links work

Who can review
--------------

@kr8n3r


kr8n3r:
@cmcnallygds  i've fixed the config for the link in the footer, but it now complains there is no H1 of the page

PR #375, opened 2020-11-05
Updated comms lead Incident Process guidance in our team manual

Jenny3020:
What
----

I've checked and updated our guidance for in hours incident process and out of hours incident process.
I've mainly updated the comms lead sections - added points about setting up hangouts, and updated the link to the incident report template (we now use a newer one). Also added the lines about updating Xgov slack paas channel. The 1st para under 'Out of hours incident process' wasn't super clear, so i've tidied that up a bit. I did this to make it easier for comms leads and support engineers to understand what they need to do in the event of an incident.

How to review
-------------

Read through as if you've been asked to handle an incident, and check that:
- you would know whats expected of you (as a comms lead)
- links work and direct you to relevant docs/guidance

Who can review
--------------

Jenny can't review, anyone else from the team can.

### DIFF
--- a/source/incident_management/incident_process.html.md.erb
+++ b/source/incident_management/incident_process.html.md.erb
@@ -28,20 +28,23 @@ Your responsibility depends on your role within the ‘incident team’:
 
 #### Incident comms lead
 1. Let the PaaS team know about the incident on #paas-incident Slack channel.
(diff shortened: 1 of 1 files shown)


kr8n3r (approved):
reads better to me. @schmie any feedback?

PR #389, opened 2021-01-26
Make guidance clearer and more concise

cmcnallygds:
What
----

I've reworded some of this guidance to make it clearer and more concise

How to review
-------------

1. Preview the content locally
2. Check that it still makes sense from a technical perspective

Who can review
--------------

not @cmcnallygds

### DIFF
--- a/source/monitoring_alerting/audit.html.md
+++ b/source/monitoring_alerting/audit.html.md
@@ -24,11 +24,13 @@ To better audit operator actions, we ship audit events to [Splunk](https://gds.s
 
 We ship events to Splunk in two ways:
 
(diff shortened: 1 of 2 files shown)


source/monitoring_alerting/audit.html.md
```diff
@@ -1,21 +1,19 @@
-# Audit
+# Auditing operator and platform events
 
 To hold operators and users to account, and to meet compliance requirements, we
 store and audit operator and platform events.
 
 ## BOSH
 
-In each environment we have a VM which runs BOSH and has components:
+In each environment, we have a virtual machine (VM) which runs [BOSH] (https://bosh.io/docs/) and has the following components:
```
kr8n3r, line 8: @cmcnallygds I think this link syntax has an extra space`[] () so` the build fails

schmie (approved):
:+1:

PR #404, opened 2021-05-26
Document AZ management

paroxp:


source/guides/managing_azs_on_paas.md
```diff
@@ -0,0 +1,29 @@
...
+Terraform, or simply running the appropriate Concourse job.
+
+[The `paas-cf` repository][1] contains a terraform module, which will deny all
+traffic, to subnets in one particular AZ.
+
+It also has a Concourse jobs, in `create-cloudfoundry` pipeline, and the
+`operator` groups, to bot disable and enable particular AZs.
+
+_Please note_: Disabling the AZ `a`, will likely bring the Concourse down,
+meaning the re-enabling action will need to be done by hand from op's machine.
```
LeePorte, line 24: Can we actually describe how to run both actions by hand as part of the guide?

paroxp: We thought this may be better suited closer to the code itself: https://github.com/alphagov/paas-cf/pull/2666/files#diff-1cb0f46e57ff170cbb3658a9d35c7c6166eb48b84b140fe7f9a69dfa8422d9b1R7

Do we want this info duplicated?

LeePorte: Yes, as if we're doing it then we're in an incident and the easier the access to the info the better

paroxp: I've added the PR, but I have doubts this will be useful :D In the incident, I'm unlikely to read the team manual myself - not a play book. Likewise, README closer to the source, will be for sure updated, whilst this can drift a little until someone remembers ¯\_(ツ)_/¯

In the same time, it doesn't matter. It's like 50 characters now...

PR #406, opened 2021-06-04
Fix "Looking up users by Google IDs" page

kr8n3r:
What
----
Fix page suffix
Currently page is not viewable


Krenair (approved):
seems to be consistent with the others, let's try it

PR #412, opened 2021-07-08
Document restoring BOSH Resurrector

paroxp:


source/guides/restoring_bosh_resurrector.md
```diff
@@ -0,0 +1,43 @@
...
+
+In addition, our [Concourse pipelines][1] will be checking up on the resurrector, and if that isn't enabled, the pipelines will fire [nagging emails][2] as a reminder.
+
+## What do I do?
+
+1. Using `paas-bootstrap`, you will need to gain access to BOSH's CLI. This can be achieved by running the following command:
+
+```
+make $ENV bosh-cli
+```g
```
AP-Hunt, line 22: Extra g

source/guides/restoring_bosh_resurrector.md
```diff
@@ -0,0 +1,43 @@
...
+
+This action requires you to be in a range of approved IPs - like a VPN.
+
+2. When your environment is ready, you will need to log into BOSH. Follow the instructions in your terminal, to obtain one time password.
+
+```
+bosh login
+```
+
+This action, will likely trigger an incident on PagerDuty if performed on production environments.
```
AP-Hunt, line 32: This is true now, do we want it to be true in the future too?

AP-Hunt (commented):
We should link to this in the nagging emails

source/guides/restoring_bosh_resurrector.md
paroxp: 👍

source/guides/restoring_bosh_resurrector.md
paroxp: Well, assuming BOSH may not be the future, then no?

I could reword it to something like:

> This action will be audited and may trigger some alerts when performed on production environments.

How does that sound? Perhaps you have something in mind?

AP-Hunt: I meant the PagerDuty incident part. I know it currently triggers in `prod` and `prod-lon` in- and out-of-hours, but is that something actually want?

paroxp:
Yeah, been planning to link soon: https://github.com/alphagov/paas-cf/pull/2701

source/guides/restoring_bosh_resurrector.md
paroxp: I know what you mean. I'm not sure. In my head people should be aware of this, in case there is either panic or an OOH call, when the "incident" wasn't expected.

But I'm also happy to remove this line and expect people to be on top of what's going on?

PR #414, opened 2021-07-16
responding to alerts: add section on cell disk usage being near full

risicle:
What
----

Add section describing actions to take if a diego cell's ephemeral disk usage is near full. We weren't able to diagnose the true cause of this because we let the issue fall out of our log retention, but the next person to encounter it would probably appreciate being told what action will make the immediate danger go away.

How to review
-------------

Read it.

Who can review
--------------

Anyone with taste.

### DIFF
--- a/source/support/responding_to_alerts.html.md
+++ b/source/support/responding_to_alerts.html.md
@@ -245,8 +245,7 @@ As this is the first monitor of this type please investigate the gorouters to di
 
 ## Diego cell ephemeral disk usage near full
 
-In the short term, to prevent a tenant-visible issue, it's probably a good idea to recreate
-the cell in bosh:
+To prevent a tenant-visible issue in the short term, it's probably a good idea to recreate the cell in bosh:
 
 ```
 cd ${PAAS_BOOTSTRAP_DIR}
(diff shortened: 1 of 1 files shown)


source/support/responding_to_alerts.html.md
```diff
@@ -242,3 +242,22 @@ If you see an alert for gorouter latency being high;
 * Contact the team who own the service to see if they are aware of anything happening
 
 As this is the first monitor of this type please investigate the gorouters to discover the issue they are encountering. We have previously seen high resource usage (CPU and Memory) these should be checked in the first case.
+
+## Diego cell ephemeral disk usage near full
+
+In the short term, to prevent a tenant-visible issue, it's probably a good idea to recreate
```
0atman, line 248:
```suggestion
To prevent a tenant-visible issue in the short term, it's probably a good idea to recreate
```
seems clearer

source/support/responding_to_alerts.html.md
```diff
@@ -242,3 +242,22 @@ If you see an alert for gorouter latency being high;
...
+the cell in bosh:
```
0atman, line 249:
```suggestion

```

0atman, line 249:
```suggestion
To prevent a tenant-visible issue in the short term, it's probably a good idea to recreate the cell in bosh:
```
seems clearer

PR #418, opened 2021-09-20
[#179620265] add ADR047 "Postgres allowed-extensions approach"

risicle:
What
----

Added ADR047 so that the next person to pick up a "release a new major postgres version to tenants" story will have a better clue what to do about extensions.

How to review
-------------

Read?

Who can review
--------------

Leadership I guess.

### DIFF
--- a/source/architecture_decision_records/ADR047-postgres-extensions-approach.html.md
+++ b/source/architecture_decision_records/ADR047-postgres-extensions-approach.html.md
@@ -1,0 +1,77 @@
+---
+title: ADR047 - Postgres allowed-extensions approach
+---
+
+# ADR047: Postgres allowed-extensions approach
+
+## Context
+
+Up to the release of postgres 13 plans to tenants, the policy for choosing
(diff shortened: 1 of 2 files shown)


source/architecture_decision_records/ADR047-postgres-extensions-approach.html.md
```diff
@@ -0,0 +1,77 @@
...
+made to https://github.com/alphagov/paas-rds-broker/blob/main/rdsbroker/supported_extensions.go .
+
+Beyond making sure an extension loads, actually trying out the functionality
+of each extension is beyond the time committment we're willing to give
+this for now. We may update the documentation to note the semi-supported
+nature of postgres extensions on our platform.
+
+## Status
+
+??
```
LeePorte, line 68: I think this should be `Accepted`

risicle: Didn't like to presume :blush:

cadmiumcat (commented):
This seems clear to me 👍

PR #435, opened 2022-04-11
[#181627481] Replace elasticsearch references with opensearch

mogds:
What
----

Replaces elasticsearch references with opensearch where appropriate 

How to review
-------------

Code review

Who can review
--------------

Not me


cadmiumcat (commented):
Nice. There's also a guide on [shipping elasticsearch metrics to tenants](https://github.com/alphagov/paas-team-manual/blob/main/source/support/shipping_elasticsearch_metrics_to_tenants.html.md) - we should remove it unless there's a good reason to keep it

AP-Hunt:
> given that they're decision records, should they be changed?

No. We should add a new ADR documenting how we changed our minds/circumstances changed and reference the old one for context.

AP-Hunt (approved):

PR #437, opened 2022-04-26
Correct guidance about expiring CDN certs

AP-Hunt:
What
----

The guidance mentioned Lets Encrypt for certificates issued on behalf of the CDN broker, but we've used AWS Certificate Manager for some time now.

How to review
-------------
Copy review


cadmiumcat (approved):
🙇

PR #438, opened 2022-05-11
Update dependencies and fix CI build

kr8n3r:
What
----

- Update dependencies.
   Fixes the issue with cached search term
- update circleCI ruby docker image
  https://discuss.circleci.com/t/legacy-convenience-image-deprecation/41034
- reset bundle cache with updated image
- Set target bundle platforms
  https://github.com/rubyjs/mini_racer/issues/227#issuecomment-1015223876
- force circleCI to install latesr bundler version before building
- update bundler `path` setting to resolved:
[DEPRECATED] The `--path` flag is deprecated because it relies on being
remembered across bundler invocations, which bundler will no longer do
in future versions. Instead please use `bundle config set path
'vendor/bundle'`, and stop using this flag

How to review
-------------

- review by commit
- can run locally to check searching
- check build succeeds
- tests pass

Who can review
--------------

not @kr8n3r


cadmiumcat (approved):
Works locally 🙌

PR #441, opened 2022-05-23
ADR: Handling AWS-based backing service maintenance

risicle:
What
----

Perhaps this shouldn't be an ADR as it is possibly more of a RFC, but we don't have a mechanism for RFCs right now. Instead here's a draft PR of the ADR to take a look at, and perhaps people can discuss whether this is a good idea here.

How to review
-------------

:eyes: 

Who can review
--------------

Humans

### DIFF
--- a/source/architecture_decision_records/ADR0050_handling_aws_backing_service_maintenance.md
+++ b/source/architecture_decision_records/ADR0050_handling_aws_backing_service_maintenance.md
@@ -1,0 +1,55 @@
+# ADR0050: Handling AWS-based backing service maintenance
+
+## Status
+
+Proposal
+
+## Context
+
+We have two types of backing service provided by AWS (RDS databases and Elasticache redis instances) that have a similar problem - when AWS need to perform maintenance on service instances we don't have a consistent, low-effort way of handling this and/or communicating this to the user. For high-availability (HA) service instances, application of these maintenance packages should result in very little downtime (perhaps a couple of minutes) as the services fail over to the standby instances. The bigger threat to uptime is from a poorly configured application failing to respect the DNS TTL and continuing to attempt connecting to the "old" IP of the service.
+
+For these services, AWS has quite a comprehensive and detailed set of features for managing these updates, but we all tend to agree that the last thing we want to do is either:
+
+ - try to expose the full power of this interface to tenants programmatically
+ - become the human proxies of these features to our tenants
+
+Both service types have a concept of a "maintenance window", a time period when most maintenance will be scheduled to take place. This time period is selectable in AWS, but we only currently expose this configurability to tenants in the RDS broker.
+
+Only very critical updates will be force-applied in this maintenance window. For less critical updates, we (the PaaS team) receive a notification that a new piece of maintenance needs to take place, and we can then make a choice as to when the maintenance occurs. For RDS, one of these choices is to perform it in the "next maintenance window". The problem with this though is we have over 800 databases in the London region alone, and the people who are best placed to know when's best to apply these updates are tenants. So we can find ourselves with a lot of communication to do to a lot of tenants, many of whom will have further questions or want boutique actions taken. In short, it's a support bomb. As a result, most of the lower priority updates sit in our "TODO" pile indefinitely and eventually get force-applied by AWS at the end of their acceptable deferral period. This is far from ideal, and is arguably as disruptive as applying all mandatory updates in the first available maintenance window, as maintenance packages will eventually be reaching the end of their deferrable period at approximately same rate as new ones appear, just **many** months late. RDS also has non-mandatory maintenance updates, which in theory can be deferred indefinitely, but these don't seem to have been common lately.
+
+There are a couple of relevant differences between RDS and Elasticache in how maintenance works:
+
+ - Elasticache doesn't have the same ability to set an item of maintenance to be performed in the next available maintenance window. Maintenance can be applied immediately, applied "automatically after the due date" (see below), or ignored (for non-mandatory updates). To achieve the same ability as RDS we'd have to run our own cron-based trigger to launch maintenance at the right time.
+ - Elasticache allows maintenance to be applied automatically in the maintenance window following an issue's "due date" (which is a date a certain number of weeks after its announcement, based on the severity of the issue). There doesn't *seem* to be an equivalent of this in RDS. As long as we're not intending to grant tenants the ability to choose to apply a specific piece of maintenance "early" at an opportune moment, it's not clear what advantage this would have for us - unless perhaps it meant lower-priority maintenance could be coalesced with higher-priority maintenance and result in less overall downtime. There's also the possibility it could be used as a substitute for the missing "apply in next maintenance window" feature.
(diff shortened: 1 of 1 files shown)


pauldougan:
I'll add the RFC idea to the retro board for the team to discuss today

pauldougan:
I'll add the RFC idea to the retro board for the team to discuss today

source/architecture_decision_records/ADR0050_handling_aws_backing_service_maintenance.md
```diff
@@ -0,0 +1,55 @@
...
+ - A flag on service instances that would allow the above "eager maintenance" feature to be turned off. A tenant might want to use this flag when they have a particularly busy or critical period approaching for their service. Using this would give them an "update holiday", which the periodic job would catch up on missed maintenance. Tenants that were extremely averse to updates could leave this flag off permanently, which would result in something close to the current situation - important updates being force-applied by AWS after the maximum deferral period.
+ This flag could also have a third setting, which would allow the tenant to only apply "mandatory" maintenance eagerly.
+ The default value for this flag is open for discussion, though leaving eager updates "off" by default would fail to solve the bulk of the problem. "Eager maintenance" is probably most useful to tenants who aren't attentive enough to discover this flag and turn it on.
+
+As a stretch goal, we could devise a mechanism to communicate upcoming maintenance to tenants, though this has some significant caveats:
```
pauldougan, line 42: perhaps an interface in PaaS admin to permit org managers to view and set maintenance windows on an org wide basis? This would need some user research I imagine.

pauldougan, line 42: perhaps an interface in PaaS admin to permit org managers to view and set maintenance windows on an org wide basis? This would need some user research I imagine.

risicle: Maybe in 2025.

source/architecture_decision_records/ADR0050_handling_aws_backing_service_maintenance.md
```diff
@@ -0,0 +1,55 @@
...
+ - A flag on service instances that would allow the above "eager maintenance" feature to be turned off. A tenant might want to use this flag when they have a particularly busy or critical period approaching for their service. Using this would give them an "update holiday", which the periodic job would catch up on missed maintenance. Tenants that were extremely averse to updates could leave this flag off permanently, which would result in something close to the current situation - important updates being force-applied by AWS after the maximum deferral period.
```
cadmiumcat, line 38: I guess this means we need to let users know the "apply by" date for each update?

cadmiumcat: There "severity"  of elasticache updates:
1. critical: Recommended to apply immediately (within 14 days or less)
2. important: Recommended to apply as soon as your business flow allows (within 30 days or less)
3. medium: Recommended to apply within 60 days or less
4. low: Recommended to apply within 90 days or less

I think the "update holiday" works with `important`, `medium` and `low` updates. But for `critical` updates, should we allow this as well?

risicle: > I guess this means we need to let users know the "apply by" date for each update?

Not necessarily. When you go on holiday you set your answerphone, but you don't necessarily know whether anyone's going to call or not.

(granted, 20 year out of date example)

risicle: As far as the severity levels go, it's not terribly well-defined how they relate to updates being "mandatory" and my tendency would be to allow AWS to be the ones to decide which ones are mandatory or not, and not make any extra stipulations of our own.

Much of the idea of this flag is to allow a tenant to opt out *entirely* of our new eager-update feature if they find it doesn't fit their needs.

cadmiumcat: Say you're planning to have a "holiday" for a month, but there is a mandatory update happening on the first week of it because that's its due date - what happens then?

wryobservations: Checking I've understood - currently we choose to apply updates, thus communicating a maintenance window, could result in some downtime, tenants react variously, huge support burden. Proposal - we automate maintenance as default giving tenants the ability to override this as and when or permanently?

source/architecture_decision_records/ADR0050_handling_aws_backing_service_maintenance.md
```diff
@@ -0,0 +1,55 @@
...
...aaS team) receive a notification that a new piece of maintenance needs to take place, and we can then make a choice as to when the maintenance occurs. For RDS, one of these choices is to perform it in the "next maintenance window". The problem with this though is we have over 800 databases in the London region alone, and the people who are best placed to know when's best to apply these updates are tenants. So we can find ourselves with a lot of communication to do to a lot of tenants, many of whom will have further questions or want boutique actions taken. In short, it's a support bomb. As a result, most of the lower priority updates sit in our "TODO" pile indefinitely and eventually get force-applied by AWS at the end of their acceptable deferral period. This is far from ideal, and is arguably as disruptive as applying all mandatory updates in the first available maintenance window, as maintenance packages will eventually be reaching the end of their deferrable period at approximately same rate as new ones appear, just **many** months late. RDS also has non-mandatory maintenance updates, which in theory can be deferred indefinitely, but these don't seem to have been common lately.
```
AP-Hunt, line 18: Do we have any idea how many times we've had emails about security updates to RDS instances? My feeling is it's not many.

AP-Hunt:
I've made some changes to make the ADR match our discussions. I'd appreciate some input on paragraphs 4 and 5 of the decision section.

source/architecture_decision_records/ADR0050_handling_aws_backing_service_maintenance.md
```diff
@@ -6,50 +6,72 @@ Proposal
...
 
- - The more tenants see about these updates, the more likely we are to get boutique support requests for special handling of their service instance and information on whether a specific update has been applied to their service.
- - If we publish this information in a general, whole-platform way, (easy for Elasticache, much harder for RDS), we leave tenants to figure out for themselves which maintenance item applies to each of their instances.
- - If we push this information to them, per-instance, we're just passing on the noise to them that we ourselves find so hard to handle. We also don't have a great way of targeting the relevant users - we tend to email org managers, which, in large organizations, can be a very blunt tool.
+Both brokers will need a mechanism for disabling maintenance updates on service instances. Given the impending decomissioning of the platform, we feel that building this out in a self-service mannner could be unnecessarily burdonsome. Having a list of identifiers in a config file which we update in response to tenant tickets would likely suffice.
```
risicle, line 65: Alternatively we could manually set a tag on the AWS resource, which would be easier to build-out to a self-service feature if we chose to. And it would mean we didn't have to include details of specific important service ids etc in our (public) paas-cf deployment details.

AP-Hunt: What do you think @alphagov/team-government-paas-people

deanwilson:
This project is archived and no longer active. Thank you for your interest.

PR #445, opened 2022-06-22
document alternate grafana's auth quirk

risicle:
What
----

https://www.pivotaltracker.com/story/show/182097389

Logging in to `grafana-2` requires the use of a little trick. Document this. I'm not sure if this is the best place for this addition, but couldn't see anywhere better immediately.

Fixing this to get rid of the quirk would be surprisingly tricky and require forking upstream's prometheus-boshrelease which we'd all rather not.

How to review
-------------

:eyes: 

Who can review
--------------

Not @risicle


cadmiumcat (commented):
😆 excellent.
Maybe linking to this in https://team-manual.cloud.service.gov.uk/technical_design/prometheus/#secondary would be handy?

cadmiumcat (commented):
this works for me. 
I'm guessing it is expected that we have no dashboards in grafana-2?

risicle:
> Maybe linking to this in https://team-manual.cloud.service.gov.uk/technical_design/prometheus/#secondary would be handy?

Now y'see that's exactly the section I didn't find. I think I might put the explanation there and link to it from the support page instead.

AP-Hunt (approved):

PR #447, opened 2022-07-15
Change CI tooling

LeePorte:
What
----

As we are coalescing around GitHub Actions and there is likely to be a fair number of tech docs template based repos that are going to be doing similar it makes sense to do this with our own first.

How to review
-------------

Check the tests pass and a code review

Who can review
--------------

Anyone


.github/workflows/pr-tests.yml
```diff
@@ -0,0 +1,39 @@
+---
+name: pr-tests
+on: pull_request
+
+env:
+  NODE_VERSION: "14.x"
+  RUBY_VERSION: "2.7.1"
```
kr8n3r, line 7: 2.7.2

.github/workflows/pr-tests.yml
```diff
@@ -0,0 +1,39 @@
...
+      - uses: ruby/setup-ruby@08245253a76fa4d1e459b7809579c62bd9eb718a #v1.111.0
+        with:
+          ruby-version: "${{env.RUBY_VERSION}}"
+          bundler-cache: true
+
+      - uses: actions/setup-python@c4e89fac7e8767b327bbad6cb4d859eda999cf08 #v4.1.0
+        with:
+          python-version: "${{env.PYTHON_VERSION}}"
+
+      - run: npm ci
```
kr8n3r, line 32: we don't have any npm dependecies on this one

.github/workflows/pr-tests.yml
```diff
@@ -0,0 +1,39 @@
...
+      - uses: actions/setup-node@2fddd8803e2f5c9604345a0b591c3020ee971a93 #v3.4.1
+        with:
+          node-version: "${{env.NODE_VERSION}}"
+
+      - uses: ruby/setup-ruby@08245253a76fa4d1e459b7809579c62bd9eb718a #v1.111.0
+        with:
+          ruby-version: "${{env.RUBY_VERSION}}"
+          bundler-cache: true
+
+      - uses: actions/setup-python@c4e89fac7e8767b327bbad6cb4d859eda999cf08 #v4.1.0
```
kr8n3r, line 28: i think python is only if you want to run locally, don't think it ran on circle CI

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,35 @@
+name: Deploy
+
+on:
+  push:
+    branches:
+      - main
+  workflow_dispatch:
+
+concurrency: production
```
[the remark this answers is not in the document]
LeePorte: `When a concurrent job or workflow is queued, if another job or workflow using the same concurrency group in the repository is in progress, the queued job or workflow will be pending.`

from https://docs.github.com/en/actions/using-jobs/using-concurrency

kr8n3r (approved):
looks good now. we 'll see the deployment in action once this is merged

PR #455, opened 2022-08-15
Notify via Slack if the deployment fails

whi-tw:
What
----

Send a slack notification if the build fails.

How to review
-------------

Code review

Who can review
--------------

Not me

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -22,7 +22,7 @@ jobs:
 
       - uses: actions/setup-node@2fddd8803e2f5c9604345a0b591c3020ee971a93 #v3.4.1
         with:
-          node-version: '14'
+          node-version: "14"
 
       - uses: ruby/setup-ruby@08245253a76fa4d1e459b7809579c62bd9eb718a #v1.111.0
         with:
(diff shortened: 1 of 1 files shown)


kr8n3r:
is there a reason we're using this and not Slack's own GH action?

AP-Hunt:
Agree with Jani. If we can use Slack's official thing then I think we ought to.

whi-tw:
@AP-Hunt @kr8n3r This action 'just works', rather than Slack's, which involves setting up all the custom message blocks.

To me, there doesn't seem to be a whole load of point in attempting to replicate the functionality of this action, having to learn the Slack message API, when there's an action which does specifically what we need.

Happy to put the time into changing the implementation if that is really what we want, though.

kr8n3r:
you should be able to just do 
`payload: "{"text": "Deployment of ${{ github.event.pull_request.html_url }} FAILED"}"`

cadmiumcat:
When you say "Slack's own thing", do you mean https://github.com/marketplace/actions/slack-send ? 
If it's as simple as @kr8n3r suggested, then I agree we ought to change it

AP-Hunt:
I would like us to use the Slack official action, please. I know it's a little rework, sorry.

AP-Hunt (dismissed):
See comment

kr8n3r (approved):
looks good. we'll see if it works

AP-Hunt (approved):

PR #463, opened 2022-11-24
Prevent search engine bots from indexing

kr8n3r:
What
----

Config option adds <meta name="robots" content="noindex"> tag in the <head>

Partial revert of https://github.com/alphagov/paas-team-manual/pull/200 as we no longer use google search so no indexing required anymore


risicle:
There _are_ multiple occasions I've been rescued by google directing me back to our own manual...

PR #464, opened 2022-11-24
[#183708862] Use Ruby 3.1.0

AP-Hunt:
What
----
Upgrades to Ruby 3.1.0 to be consistent with the rest of our estate

How to review
-------------

1. Checks pass
2. `bundle exec middleman`

Who can review
--------------
@kr8n3r would be valuable


kr8n3r:
@AP-Hunt this is unlikely going to work as work most likely needs to be done upstream in the tech-docs-gem  (if packages actually support ruby 3)

AP-Hunt:
Bugger. I ran `bundle exec middleman` locally and it worked. Maybe `rbenv` had me using the wrong Ruby version.

kr8n3r:
https://github.com/gjtorikian/html-proofer/issues/781 error given is fixed in v5 but v5 is not available in GH actions

AP-Hunt:
Maybe we can check back in on this in a week or two? Looks like those whole thing is still active.

kr8n3r:
@AP-Hunt all fixed. Can you review changes and squash commits if happy?

AP-Hunt:
I can't review because I opened the PR, but I'm happy with what's there. I don't think the commits need squashing either, they make sense in isolation.

kr8n3r (approved):

PR #466, opened 2023-01-06
ADR-050 cloudapps domain future

LeePorte:
What
----

Added in ADR-050 which lays out what we are going to do with the cloudapps.digital domain post the closure of the platform. 

I have also removed the strikethrough attempts on ADR-048 and ADR-049 as these do not render correctly.

How to review
-------------

Read through the ADR and see if it makes sense. feel free to add comments and counter points. Check that it all renders correctly

Who can review
--------------

Anyone

### DIFF
--- a/source/architecture_decision_records/ADR050-cloudapps-digital-domain.html.md
+++ b/source/architecture_decision_records/ADR050-cloudapps-digital-domain.html.md
@@ -1,0 +1,35 @@
+---
+title: ADR050 - Plans for the cloudapps.digital domain post platform retirement
+---
+
+# ADR0050: Plans for the cloudapps.digital domain post platform retirement
+
+## Context
+
+GOV.UK PaaS has provided a [default domain for applications](https://docs.cloud.service.gov.uk/orgs_spaces_users.html#regions). It has always been part of the documentation and [using a custom domain](https://docs.cloud.service.gov.uk/deploying_services/use_a_custom_domain/) has been supported since the platforms inception.
+
+The intention of the cloudapps.digital domain was that it gave tenants whilst in the development stages a way to quickly gain feedback on if the platform was right for their use case, and that live applications should not be hosted on it.
+
+We are aware that some tenants have used this domain for non development purposes
+
+The technical method of how subdomains of cloudapps.digital have been registered is that the wildcard for cloudapps.digital points to GOV.UK PaaS. On the PaaS [route registrar](https://github.com/cloudfoundry/route-registrar) is a running daemon that accepts requests from the Cloudfoundry API to broadcast routes to [gorouter](https://github.com/cloudfoundry/gorouter) to match inbound traffic to subdomains of cloudapps.digital and to custom domains.
+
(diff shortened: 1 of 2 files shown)


source/architecture_decision_records/ADR050-cloudapps-digital-domain.html.md
```diff
@@ -0,0 +1,35 @@
...
+
+The intention of the cloudapps.digital domain was that it gave tenants whilst in the development stages a way to quickly gain feedback on if the platform was right for their use case, and that live applications should not be hosted on it.
+
+We are aware that some tenants have used this domain for non development purposes
+
+The technical method of how subdomains of cloudapps.digital have been registered is that the wildcard for cloudapps.digital points to GOV.UK PaaS. On the PaaS [route registrar](https://github.com/cloudfoundry/route-registrar) is a running daemon that accepts requests from the Cloudfoundry API to broadcast routes to [gorouter](https://github.com/cloudfoundry/gorouter) to match inbound traffic to subdomains of cloudapps.digital and to custom domains.
+
+As this method is part of the platform itself, there is no standalone method of maintaining domain routing information. Once the platform is decommissioned this will no longer function.
```
stephengrier, line 17: But the DNS for `cloudapps.digital` is managed in a Route53 zone though so will/could live on after CloudFoundry is turned off.

It would be fairly trivial I think to add extra A records to the Route53 zone to resolve specific PaaS routes to GOV.UK's bouncer service to 301 redirect to a new hostname. So for example, you'd have:

```
*.cloudapps.digital.  CNAME. prod-cf-rtr-apps-1104940482.eu-west-1.elb.amazonaws.com.
myapp.cloudapps.digital.  CNAME  redirector-cdn.production.govuk.service.gov.uk.
```

Something like this might make it easier for folks to migrate off the platform. They'd still need a new domain for their apps but they would at least have a redirect from their old hostnames.

LeePorte: There are in the region of 7,800 subdomains in use on the cloudapps.digital domain. It's a question of managing it that is the core of the issue.

cadmiumcat: By the time the platform has been decommissioned, we expect to have removed support, so I don't think it would be beneficial to allow tenants to continue to depend on us in that way for longer. 

They already have a way of setting up a redirect once they have purchased a domain

pauldougan: Personally I don't think it's worth using bouncer since no tenants should not be using a cloudapps.digital style address for production services. Plus it would just give someone some work to maintain this it does not sound like a good use of the teams time.

Tenants are at liberty to host an app that redirects to their new service but obviously this is dependant on the paas platform still running, after dec 23 2023 the platform will not run and any self hosted redirect would go away. 

Perhaps we should make this super clear in the [cdn section of the migration guidance](https://www.cloud.service.gov.uk/migration-guidance/#paas-cdn-migration)?

PR #470, opened 2023-01-27
Close billable accounts: Signpost to suspension script in paas-cf

whi-tw:
## What

Signpost to script for suspending accounts in paas-cf

script was added in alphagov/paas-cf#3073, so this will need to be merged first for this change to be correct

## How to review

👀

## Who can review

Not me

### DIFF
--- a/source/accounts/close_billable_accounts.html.md
+++ b/source/accounts/close_billable_accounts.html.md
@@ -20,7 +20,7 @@ A closed account:
 
 1. Tenant (an org manager or billing manager) confirms that they are ready to close their org account with us.
 2. Support engineer performs all technical decommissioning tasks.
- - Suspend org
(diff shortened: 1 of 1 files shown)


source/accounts/close_billable_accounts.html.md
```diff
@@ -20,7 +20,7 @@ A closed account:
 
 1. Tenant (an org manager or billing manager) confirms that they are ready to close their org account with us.
 2. Support engineer performs all technical decommissioning tasks.
- - Suspend org
+ - Suspend org (`scripts/suspend-org.sh ${ORG_NAME}` in `paas-cf`)
```
corlettb, line 23: As andy mentioned should we also refer to the admin page process as well?

PR #489, opened 2023-06-15
[#185352990] Provide guidance on enabling private backing service access

AP-Hunt:
What
----

Whoever is on support will need to know who we are offering private backing service access to, what to ask of them, and how to enable it.

How to review
-------------
Do the instructions make sense?
Have I missed anything?
Does it look right when run?

Useful to know
------

If you can't get Ruby 3.1.1 running locally, try

```
docker run -it -v "$(pwd):/app" -p4567:4557 ruby:3.1.1 /bin/bash
cd /app
bundle install
bunde exec middleman
```

### DIFF
--- a/source/guides/vpc_peer_backing_service_access.html.md
+++ b/source/guides/vpc_peer_backing_service_access.html.md
@@ -1,0 +1,61 @@
+---
+title: VPC Peer Backing Service Accesss
+---
+
+# Enable access to backing services via VPC peering
+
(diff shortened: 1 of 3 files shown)


whi-tw (approved):
From what I can tell, this reads and runs correctly.

Does it need 2i?

AP-Hunt:
I don't think so, because this isn't changing user behaviour (unless you argue that _we_ are users of our thing .. which .. like,  yeah that's a fair argument)

PR #579, opened 2024-09-19
Add steps for decommissioning London production env

monotypical:
What
----

Add steps for decommissioning London production env.

How to review
-------------

Check out the branch, run `bundle exec middleman server` (after `bundle install`ing the dependencies) and browse the generated site locally using the URL given in the console output


DominicGriffin:
These are all the same for Ireland. Suggest we combine them/find a way to make less redundancy so we can edit easier

